PR #382, opened 2014-12-08
[Translator] Feature/translator

mmoreram:


src/Elcodi/Component/Core/Generator/UniqIdGenerator.php
```diff
@@ -0,0 +1,37 @@
...
+class UniqIdGenerator implements GeneratorInterface
+{
+    /**
+     * Generates a unique Id
+     *
+     * @return string Result of generation
+     */
+    public function generate()
+    {
+        return rand(0, 2147483647);
```
mickaelandrieu: why not a md5(md5(uniqid())) ?

mmoreram: md5(md5(uniqid())) is easily findable. This is actually an old issue :) Take a look at the actual implementation.

PR #450, opened 2015-01-07
Fixed incorrect doctrine version in composer.json

alch:
Fixed incorrect version constraint "~2.6" in configuration bundle


xphere:
I think this is a "replace all" bug when moving towards symfony 2.6 :+1:

mmoreram:
Absolutely... :)

Thanks @alch!

PR #451, opened 2015-01-07
[Configuration] Add `elcodi_config` DI function

### LINKED ISSUE #448: Register `elcodi_config` in ExpressionLanguage
With the new [extension providers](http://symfony.com/doc/current/components/expression_language/extending.html#using-expression-providers) of ExpressionLanguage, we can register our own `elcodi_config` function and use this as a replacement for `@=service("elcodi.configuration_manager").getParameter('namespace.key')`, so we can only write `@=elcodi_config('namespace.key')` instead.

xphere:
Fixes #448 
Depends on symfony v2.6.2

*UPDATE:* Depends on 2.6.3, since 2.6.2 has security issues.

### DIFF
--- a/src/Elcodi/Bundle/ConfigurationBundle/ElcodiConfigurationBundle.php
+++ b/src/Elcodi/Bundle/ConfigurationBundle/ElcodiConfigurationBundle.php
@@ -16,6 +16,7 @@
 
 namespace Elcodi\Bundle\ConfigurationBundle;
 
+use Elcodi\Component\Configuration\ExpressionLanguage\ConfigurationExpressionLanguageProvider;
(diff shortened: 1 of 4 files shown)


mmoreram:
Great ! :dancer:

alch:
great! I would also change the parameter injection in service definition for tests in https://github.com/elcodi/elcodi/blob/master/src/Elcodi/Bundle/ConfigurationBundle/Tests/Functional/app/config.yml#L53-L67. 

We will implicitly make sure that the correct symfony 2.6.2 version is installed

xphere:
Done! Thanks @alch

PR #457, opened 2015-01-09
Added cache_prefix in BambooBundle

mmoreram:


xphere:
Simple and elegant :necktie: :+1:

PR #460, opened 2015-01-09
Fixed translator

mmoreram:
* When a field is not found in the Form, this field is not managed
* Fixed the way "required" flag is loaded


src/Elcodi/Component/EntityTranslator/Form/Type/TranslatableFieldType.php
```diff
@@ -168,10 +168,12 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 : '';
 
             $builder->add($translatedFieldName, $fieldType, array(
-                'required' => $this->evaluateRequired(
-                    $fieldOptions['required'],
-                    $locale
-                ),
+                'required' => $fieldOptions['required']
```
xphere: if this is not always `bool`, maybe better check with `isset`. If it is, `evaluateRequired` will always receive `true` as first parameter.

mmoreram: ñee, true...

src/Elcodi/Component/EntityTranslator/Form/Type/TranslatableFieldType.php
```diff
@@ -189,9 +191,9 @@ public function buildForm(FormBuilderInterface $builder, array $options)
      */
     public function evaluateRequired($required, $locale)
     {
-        return $this->fallback
-            ? $this->masterLocale === $locale
-            : $required;
+        return (boolean) $required
```
xphere: This tertiary looks quite hard to read, don't know if we can improve this one with `if`.

PR #461, opened 2015-01-09
Added format to image urls

mmoreram:
* Also changed controller where these routes are generated
* Also changed twig extensions


src/Elcodi/Bundle/MediaBundle/DependencyInjection/Configuration.php
```diff
@@ -101,7 +101,7 @@ protected function setupTree(ArrayNodeDefinition $rootNode)
                                     ->defaultValue('elcodi_media_image_resize')
                                 ->end()
                                 ->scalarNode('controller_route')
-                                    ->defaultValue('/image/{id}/resize/{height}/{width}/{type}')
+                                    ->defaultValue('/image/{id}/resize/{height}/{width}/{type}.{_format}')
```
xphere: We need to allow format whitelisting, adding requirements to `_format` for this route. Maybe we can add a parameter?

PR #462, opened 2015-01-09
Added some phpdoc

mmoreram:
@xphere Attention please.


xphere:
:+1: Sure thing, my fault.

This should've emerged before accepting PR. I think we've been doing this too hastily and we need to put more "love" in our reviews. Thanks for the fix.

PR #470, opened 2015-01-14
Sorted languages in twig function

mmoreram:
* Master languages must be always the first one
* New returns array


mmoreram:
Related with https://github.com/elcodi/bamboo-admin/issues/83

PR #471, opened 2015-01-15
Added images in manufacturer and variant

mmoreram:
* Added imagesSort in all entities with images as well
* Fixed some phpdoc
* Fixed entities, factories and mapping data

### DIFF
--- a/src/Elcodi/Component/Media/Entity/Traits/ImagesContainerTrait.php
+++ b/src/Elcodi/Component/Media/Entity/Traits/ImagesContainerTrait.php
@@ -29,6 +29,13 @@ trait ImagesContainerTrait
     protected $images;
 
     /**
+     * @var string
+     *
+     * Images sort
+     */
+    protected $imagesSort;
+
+    /**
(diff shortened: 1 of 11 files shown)


src/Elcodi/Component/Media/Entity/Traits/ImagesContainerTrait.php
```diff
@@ -67,6 +74,39 @@ public function getImages()
     }
 
     /**
+     * Get sorted images
+     *
+     * @return \Doctrine\Common\Collections\ArrayCollection
+     */
+    public function getSortedImages()
```
xphere: What's the rationale behind this one? I found the code is not easy enough to follow.
Also, is this a trait for entities? I thought no logic could go inside entites, only getters and setters.

mmoreram:
This is no business logic. There is a field called imagesSort, where all the image ids, separated by commas, and this method is just a getter of images but sorted, given imagesSort field data. In that case makes sense this field to be here.

PR #472, opened 2015-01-16
Setted Page instance translatable in Bamboo

mmoreram:
* Also fixed Interface. We should add as well the setters if the fields
* Improved the exception of translatable builder

### DIFF
--- a/src/Elcodi/Bundle/BambooBundle/Resources/config/translations.yml
+++ b/src/Elcodi/Bundle/BambooBundle/Resources/config/translations.yml
@@ -31,3 +31,12 @@ elcodi_entity_translator:
                 metaTitle: ~
                 metaDescription: ~
                 metaKeywords: ~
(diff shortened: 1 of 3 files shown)


src/Elcodi/Bundle/BambooBundle/Resources/config/translations.yml
```diff
@@ -31,3 +31,12 @@ elcodi_entity_translator:
                 metaTitle: ~
                 metaDescription: ~
                 metaKeywords: ~
+        Elcodi\Component\Page\Entity\Interfaces\PageInterface:
+            alias: image
```
xphere: C&P error? This should be `alias: page` if I'm not wrong.

mmoreram: yes, sure, thanks

PR #474, opened 2015-01-16
Fixed code

mmoreram:


xphere:
More like a CS fix, isn't it? I was trying to find the code fix between all those spaces and finally got the point. My head is slow with this disease still.

mmoreram:
Fixed, thanks.

PR #477, opened 2015-01-18
Rules reworked

xphere:
Just made a eXtended Proof of Concept to simplify the usage of rules:

- `Rule` is now the only entity in the component. No `RuleGroup`, no `Expression`.
- Added new method in `ExpressionLanguage` to evaluate other rules, so you can composite easily.
- Simplification on both `CompilerPass`.
- Tested.


xphere:
:dancer:

PR #488, opened 2015-01-23
Added configuration option for domain sharding in media

alch:
Adding configuration option to make `ImageExtension` work smooth with reverse proxies.
It will work in a similar fashion as [`assets_base_url`](http://symfony.com/doc/current/reference/configuration/framework.html#assets-base-urls) in `FrameWorkBundle`.

Example:

```
elcodi_media:
    images:
        domain_sharding:
            enabled:              true
            base_urls:            [ 'media1.elcodi.com', 'media2.elcodi.com', 'media3.elcodi.com' ]
```

The `resize` and `view` twig functions provided by the `ImageExtension` class will use the sharding configuration to change the hostname for the generated images routes

### DIFF
--- a/src/Elcodi/Component/Media/Twig/ImageExtension.php
+++ b/src/Elcodi/Component/Media/Twig/ImageExtension.php
@@ -49,21 +49,56 @@ class ImageExtension extends Twig_Extension
     protected $imageViewControllerRouteName;
 
     /**
+     * @var boolean
+     *
+     * Enable domain sharding
+     */
+    protected $imageDomainShardingEnabled;
+
+    /**
+     * @var boolean
+     *
+     * Generate absolute or relative path in URLs
+     */
+    protected $generateAbsolutePath;
+
+    /**
+     * @var array
+     *
+     * Base URLs for the resource origin
+     */
+    protected $imageDomainShardingBaseUrls;
+
+    /**
      * Construct method
      *
      * @param UrlGeneratorInterface $router                         Router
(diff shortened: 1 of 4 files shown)


xphere:
Beware with using random to choose the domain, because now we're opening the chance of avoiding local browser and proxy caches if we generate different urls for same resources...

src/Elcodi/Component/Media/Twig/ImageExtension.php
```diff
@@ -109,12 +150,41 @@ public function resize(ImageInterface $imageMedia, $options)
...
+     * in URLs can optimize browser load times by parallelizing resource
+     * download (domain sharding)
+     *
+     * @link http://www.stevesouders.com/blog/2009/05/12/sharding-dominant-domains/
+     *
+     * @return string
+     */
+    protected function randomizeBaseUrls($path)
+    {
+        if ($this->generateAbsolutePath) return $path;
```
mmoreram: Have you applied the php-cs-fixer? :(

alch: Yep. For the `{}` you mean? Fixing it now

alch:
@xphere this is ideally used with a CDN that has a single origin and several CNAMES pointing to it.
Using parallelism (domain sharding) is a common practice although people are starting to drop it because of [SPDY](http://www.chromium.org/spdy).
However, even without origin aliasing, we need some kind of twig helper that can substitute the host part in a URL if we want to support any CDN

PR #490, opened 2015-01-25
Added redis dependency in Metric and Elcodi

mmoreram:


hacfi:
For an open source library I can’t recommend a dependency on ext-redis because some users might not be able to run this on managed hosting. You should try to find a solution where is falls back to [predis](https://github.com/nrk/predis)

PR #493, opened 2015-01-26
Persistent pages

rogergros:
Added the possibility to add "persistent pages" that can be modified but not removed neither disabled.
- New class property to mark the pages as persistent

Ping @elcodi/owners


mmoreram:
Thanks @rogergros 

Feel free to remove the branch :)

PR #494, opened 2015-01-26
Add Rule logic to Coupon

xphere:


xphere:
Changed description for better discussion.

PR #503, opened 2015-02-02
Remove checking `hook_system`

### LINKED ISSUE #499: [Plugin] Can't override `elcodi.hook_system`
When I set `elcodi.hook_system` in `config.yml` to point to a service not in the `PluginBundle`, the `Extension::postLoad` method throws because it can find the service on the container. This is basically because it is working with it's own container instance, before merging.

We can solve this in two ways:
- Remove the check for service existence, like what has been done in other `postLoad` methods, and let the `DependencyInjector` to fail when instancing the service. This is more error-prone, in my opinion.
- Add a method like `Extension::onCompile` which would be called like a `CompilerPass` and allow some work to be done. This allows for better errors to be thrown and for more interesting logic too. Also, we can remove some basic compiler-passes and do their work on the `Extension` itself.

Any advice, @elcodi/owners?

xphere:
Removes the check, because the service resolution automatically detects this. (and the current approach was buggy)

Fix #499

### DIFF
--- a/src/Elcodi/Bundle/PluginBundle/DependencyInjection/ElcodiPluginExtension.php
+++ b/src/Elcodi/Bundle/PluginBundle/DependencyInjection/ElcodiPluginExtension.php
@@ -90,15 +90,6 @@ class ElcodiPluginExtension extends AbstractExtension
     {
         parent::postLoad($config, $container);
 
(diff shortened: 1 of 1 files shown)


mmoreram:
Go! :dancer: 

Thanks @xphere

PR #509, opened 2015-02-10
Using right populator adapter instead of dummy adapter

rogergros:
Using right populator on elcodi insead of the dummy adapter


rogergros:
This was already talked with @mmoreram Merging :)

PR #516, opened 2015-02-13
Task/customer add address

rogergros:
New LiteAddress Entity


src/Elcodi/Bundle/GeoBundle/Resources/config/factories.yml
```diff
@@ -15,6 +15,14 @@ services:
     elcodi.factory.address:
         alias: elcodi.core.geo.factory.address
 
+    elcodi.core.geo.factory.lite_address:
```
mmoreram: Remove the long syntax please. We must use the short syntax until now :)

PR #519, opened 2015-02-17
Fix dependencies for --prefer-lowest

xphere:
The `symfony/dependency-injection` was missing in `composer.json`.
This was installing `v2.4.0`, but we need `v2.6.3` at least (for [this issue](https://github.com/symfony/symfony/issues/13267)).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,26 @@
 language: php
 
+env:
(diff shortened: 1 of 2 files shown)


xphere:
Added a travis environment for testing with `--prefer-lowest` in PHP 5.4

PR #520, opened 2015-02-17
fix create coupon factory

jromero86:


mmoreram:
Thanks @jromero86 but we've already changed this behavior. This afternoon these changes will be merged to master :)

PR #527, opened 2015-02-19
Fix page template

xphere:
Made `TemplatedPageRenderer` as independent as possible and moved it to the `Page` component.
Now it searches for a template in various bundles, returning the first one that exists.

In Bamboo, page template is now customizable via store configuration.


xphere:
This PR and related ones will be merged today if no issues arise. /ping @elcodi/owners

mmoreram:
@xphere needs to be rebased. ASAP is rebased, will be merged.

PR #528, opened 2015-02-20
Removed unmapped inversedby 'children' property

alch:


mmoreram:
Indeed this change would make all the system fail, because we need `Comment::getChildren()` method in order to create a tree.

The error comes from the doctrine validation `php app/console doctrine:schema:validate`

``` shell
[Mapping]  FAIL - The entity-class 'Elcodi\Component\Comment\Entity\Comment' mapping is invalid:
```

And the right fix is to create this inversion definition. I have made the PR in https://github.com/elcodi/elcodi/pull/529

PR #533, opened 2015-02-23
Makes default currency writable

xphere:
Default currency is marked as `read-only` and can't be changed from services.

This also brings another problem: when an ajax request fails (like this one did, sending back a 500 HTTP error), the admin should warn the user, but now it makes him wait forever. Ping @tonipinel


rogergros:
:+1:

PR #539, opened 2015-02-25
Feature/store wizard plugin

rogergros:
Wizard plugin

TODO: Update the Wizard Status service to check the right way if an address  and the payment info has been added.

Ping @elcodi/owners


src/Elcodi/Plugin/StoreSetupWizardBundle/Controller/WizardController.php
```diff
@@ -0,0 +1,60 @@
...
+     * )
+     * @Template()
+     */
+    public function viewAction()
+    {
+        $wizardStatusService = $this
+            ->get('elcodi_templates.wizard_status.service');
+        $stepsFinished       = $wizardStatusService->getStepsFinishStatus();
+
+        $activeStep = null;
```
xphere: Wouldn't be better a `getNextStep` method in `WizardStatus` service?

src/Elcodi/Plugin/StoreSetupWizardBundle/EventListener/DashboardRedirectionEventListener.php
```diff
@@ -0,0 +1,118 @@
...
+    protected $wizardStatus;
+
+    /**
+     * Builds a new class
+     *
+     * @param Router       $router       A router
+     * @param WizardStatus $wizardStatus A wizard status service
+     */
+    public function __construct(
+        Router $router,
```
xphere: Since you're only using `generator` method, better typehint for `UrlGeneratorInterface` instead of the full `Router`.

src/Elcodi/Plugin/StoreSetupWizardBundle/Templating/TwigRenderer.php
```diff
@@ -0,0 +1,101 @@
...
+     * @param EventInterface $event The event
+     */
+    public function renderMiniWizard(EventInterface $event)
+    {
+        if ($this->plugin->isEnabled()) {
+            $stepsFinished = $this
+                ->wizardStatus
+                ->getStepsFinishStatus();
+
+            $activeStep = null;
```
xphere: Same here, this would benefit from a `getNextStep` method in `WizardStatus`.

src/Elcodi/Plugin/StoreSetupWizardBundle/Services/WizardStatus.php
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /**
+     * Checks if the wizard has already been finished
+     *
+     * @return bool
+     */
+    public function isWizardFinished()
+    {
+        $stepsFinishedStatus = $this->getStepsFinishStatus();
```
xphere: This will be `return $this->getNextStep() === null;` if we had that method.

src/Elcodi/Plugin/StoreSetupWizardBundle/Controller/WizardController.php
rogergros: Yes, for sure. I'll change it, I was planning it as I did with "isWizardFinished"

src/Elcodi/Plugin/StoreSetupWizardBundle/EventListener/DashboardRedirectionEventListener.php
rogergros: :+1: I'll change it

src/Elcodi/Plugin/StoreSetupWizardBundle/Services/WizardStatus.php
rogergros: :+1:

xphere:
:+1: Nice to have some plugins on the run.

We need to discuss a better approach for plugins activation/deactivation (even self-deactivation) and services registered from plugins should be activated/deactivated too automatically. I think the way it is now is too complex and error-prone. (needs to pass current plugin with `@=elcodi_plugin("plugin_name")` and check it to be enabled)

rogergros:
Hey @elcodi/owners this will be merged today at noon if no problem arises :)

PR #541, opened 2015-02-25
added removeProduct function in CartManager

claretcrab:
I think it's usefull if you don't know the cartLine ID but you know product ID

### DIFF
--- a/src/Elcodi/Component/Cart/Services/CartManager.php
+++ b/src/Elcodi/Component/Cart/Services/CartManager.php
@@ -409,4 +409,50 @@ class CartManager
 
         return $this;
     }
+
+    /**
+     * Remove a Purchasable from Cart
+     *
+     * This method removes a Purchasable from the Cart.
+     *
+     * If the Purchasable is already in the Cart, it just decreases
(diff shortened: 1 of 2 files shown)


alch:
Why not change `addProduct` so that it can accept a *negative* value (the inner method `Cartmanager::setCartLineQuantity` already checks that the final quantity is > 0 and if not it *removes* the product from the cart) ?

Then the `CartManager::removeProduct` would be something like

```php
public function removeProduct(
        CartInterface $cart,
        PurchasableInterface $purchasable,
        $quantity
) {
    return $this->addProduct($cart, $purchasable, -$quantity);
}
```

Opinions?

xphere:
I found quite counterintuitive that `addProduct` accepts a negative amount in quantity…
I'd rather move the `<= 0` check to `removeProduct` and throw when `addProduct` receives negative quantity.

claretcrab:
Mmmm i think it's better to have different methods for this purpose, because in addProduct there is the logic of addLine too.

claretcrab:
unit test added

alch:
:+1:

PR #563, opened 2015-02-28
Page improvements and Disqus Plugin

mmoreram:
* Created Disqus plugin for disqus comments in Blog post page and product page
* User can enable/disable comments in any page
* Improved Page Component and Bundle
* Improved Plugin helpers to get info about if a plugin is enabled or not
* Fixed all plugins for using this new method
* Some fixes


src/Elcodi/Component/Page/ElcodiPageTypes.php
```diff
@@ -0,0 +1,38 @@
...
+     * Used in footer
+     */
+    const TYPE_REGULAR = 1;
+
+    /**
+     * Page type blog post
+     *
+     * Used in blog
+     */
+    const TYPE_BLOG_POST = 2;
```
xphere: I'd rather have would go with a string `group` field, so you can request all of the pages related to a group, and create your own groups (also, this would make collecting urls easier). But I think that's something we can do in the beta time.

src/Elcodi/Component/Page/Repository/PageRepository.php
```diff
@@ -29,30 +28,70 @@
...
+     *
+     * @param string  $type          Type
+     * @param integer $page          Page
+     * @param integer $numberPerPage Number per page
+     *
+     * @return array Pages
+     */
+    public function findPages($type, $page, $numberPerPage)
+    {
+        $offset = (($page - 1) * $numberPerPage) + 1;
```
xphere: This is something the paginator should do, not the repository. Same for the next function… `getNumberOfEnabledPages`? Really?

src/Elcodi/Component/Plugin/Templating/Traits/TemplatingTrait.php
```diff
@@ -68,4 +70,28 @@ protected function appendTemplate(
...
+    protected function pluginCanBeUsed(
+        Plugin $plugin,
+        array $checkableFields = []
+    ) {
+        return array_reduce(
+            $checkableFields,
+            function ($canBeUsed, $checkableField) use ($plugin) {
+
+                return $canBeUsed &&
+                    isset($plugin->getConfiguration()[$checkableField]) &&
```
xphere: This is calling aloud for a `getSettings()` and a `getSetting($name)` methods in `PluginInterface`.

PR #567, opened 2015-03-01
Fixed wizard bug when the wizard is finished

rogergros:
The mini wizard to go to the next page was shown when the wizard was already finished


xphere:
:+1:

PR #571, opened 2015-03-01
Using fully qualified classname in trait

alch:


mmoreram:
@alch Epic thanks :dancer: :+1:

PR #583, opened 2015-03-04
Message on settings to disable under construction mode

rogergros:
Added a message to remember you that the "under construction" mode should be disabled when the wizard is finished and the user visits the settings page.

Ping @elcodi/owners


rogergros:
Rebased and ready to be merged

PR #584, opened 2015-03-04
New field active urls

rogergros:
New field "active urls" for the menu nodes. This field is used to mark the urls where this menu should be showed as "active".

ping @elcodi/owners


xphere:
Isn't this something like a related category?

rogergros:
My fault, these are routes, bad names for the test, changing it :)

PR #587, opened 2015-03-06
added display none to tracking pixels

claretcrab:


xphere:
I'm not sure this makes all browsers happy, since some may choose not to download hidden images.
If tracking pixels break your flow, I'd suggest to make them `position: absolute` just to make sure they don't take any space in the layout. Or we can ping @tonipinel.

claretcrab:
Sure, it's just only to report the bug. We can discuss with Toni to find the best solution ;)

Thanks!

javiereguiluz:
Here are some recommendations about *pixel tracking* made by Google in its [Doubleclick for Publishers documentation](https://support.google.com/dfp_premium/answer/1347585?hl=en):

```html
<img src="TRACKING-PIXEL-URL-GOES-HERE" style="position:absolute; visibility:hidden">

<img src="TRACKING-PIXEL-URL-GOES-HERE" style="display:none">

<img src="TRACKING-PIXEL-URL-GOES-HERE" width="0" height="0">
```

mmoreram:
Good :D Thanks for this :dancer: 

@tonipinel anything to say about that?

tonipinel:
I think the position absolute is better solution... with display none maybe some browsers will not load the image... and the tracking will not be saved.

PR #606, opened 2015-03-15
Twitter & Facebook plugin

rogergros:
- Twitter plugin
  - You can tweet about a product from list or from product page
  - You can tweet about a product from the admin product management
  - You can follow the twitter user from footer
- Facebook plugin
  - You can post about a product from list or from product page
  - You can post about a product from the admin product management
  - You can follow the facebook user from footer

Ping @elcodi/owners

### DIFF
--- a/src/Elcodi/Plugin/FacebookBundle/Resources/translations/messages.es.yml
+++ b/src/Elcodi/Plugin/FacebookBundle/Resources/translations/messages.es.yml
@@ -1,0 +1,15 @@
+facebook_plugin:
+    title: Facebook
+    field:
+        facebook_account:
+            title: Cuenta de facebook
+    action:
+        visit_google: Ir a Facebook
+        result_ok: Configuración de facebook guardada
+    section:
+        settings:
+            title: Configuración de Facebook
+    tweet:
+        take_look: Echa un vistazo a
+    error:
+        is_disabled: El plugin está desactivado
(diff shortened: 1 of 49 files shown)


src/Elcodi/Plugin/FacebookBundle/Resources/translations/messages.es.yml
```diff
@@ -0,0 +1,15 @@
+facebook_plugin:
+    title: Facebook
+    field:
+        facebook_account:
+            title: Cuenta de facebook
```
javiereguiluz: In this translation file you mix `facebook` and `Facebook` indistinctively. In the original `en` translation file, you always use `Facebook`. I think that's a good idea to be more consistent.

src/Elcodi/Plugin/TwitterBundle/Resources/views/Tweet/share.html.twig
```diff
@@ -0,0 +1,9 @@
+<a
+    href="https://twitter.com/share"
+    class="twitter-share-button"
+    data-url="{{ url }}"
+    data-text="{{ text }}"
```
javiereguiluz: In case the content of this `text` variable is not safely escaped, consider using [the escape Twig filter](http://twig.sensiolabs.org/doc/filters/escape.html) as follows:

```twig
data-text="{{ text|escape('html_attr') }}"
```

PR #610, opened 2015-03-16
Prevent fixtures command from always returning 1

hacfi:
I was irritated why the fixtures seemed to fail.

### DIFF
--- a/src/Elcodi/Bundle/FixturesBoosterBundle/Command/LoadDataFixturesDoctrineCommand.php
+++ b/src/Elcodi/Bundle/FixturesBoosterBundle/Command/LoadDataFixturesDoctrineCommand.php
@@ -83,13 +83,11 @@ class LoadDataFixturesDoctrineCommand extends OriginalCommand
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         if (empty($this->databaseFilePath)) {
-            parent::execute($input, $output);
-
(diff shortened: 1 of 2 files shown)


src/Elcodi/Bundle/FixturesBoosterBundle/Command/LoadDataFixturesDoctrineCommand.php
```diff
@@ -83,13 +83,11 @@ public function __construct(
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         if (empty($this->databaseFilePath)) {
-            parent::execute($input, $output);
-
-            return 1;
+            return parent::execute($input, $output);
```
xphere: So, that's why my refresh script was failing without errors after fixture loading. Thanks for the fix!
Just a minor thing, CS requires new line before return statements.

hacfi: Even if the code is indented? Ok..will update the PR

alch:
@hacfi :+1: This problem was bugging us too

xphere:
Merged in https://github.com/elcodi/elcodi/commit/a9176cc19d8d39062d963ff37c6564b8647ad848
Thanks @hacfi

PR #612, opened 2015-03-16
Remove Symfony @api references

hacfi:


src/Elcodi/Bundle/AttributeBundle/CompilerPass/MappingCompilerPass.php
```diff
@@ -26,11 +26,7 @@
 class MappingCompilerPass extends AbstractMappingCompilerPass
 {
     /**
-     * You can modify the container here before it is dumped to PHP code.
```
mmoreram: Why are you removing this block?

hacfi: Because it was just copied from https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/DependencyInjection/Compiler/CompilerPassInterface.php#L26 and here it doesn’t represent any meaningful information. Should I readd it?

mmoreram: Even if it is copied from the other method, this text is the textual representation of what this method is intended for, right?

hacfi: I'll readd it later today.

In my opinion that description wasn't helping much for developers who are
learning Elcodi as it refers to default Symfony Dependency Injection
functionality.

mmoreram:
I close this PR because is letting the docblock with unformatted.

Then, we should discuss about the format of the docblocks?

hacfi:
@mmoreram Didn’t have my laptop for a couple of days. Do you want a PR for the `@api` removed or will you discuss this first?

PR #613, opened 2015-03-16
Add use statements in Traits

hacfi:


xphere:
I remember hearing at the office that use statements in traits were giving trouble to someone's IDE, can't remember the rationale, but @alch changed this to FQCN for a good reason in commits like https://github.com/elcodi/elcodi/commit/d1fc8f64f14f1c0ec8d005021c3d5eaf158ac9c4. Maybe he can shed some light on this so we're not running in circles.

OTOH, moving from `\Doctrine\Common\Collections\Collection` to `ArrayCollection` is too specific and can be troublesome if somebody uses a `PersistentCollection` or some custom one.

hacfi:
Ok..we can close this then.

I didn’t change Collection to ArrayCollection..just adapted added the use statement. We need a separate PR to address the usage of ArrayCollection.

alch:
Yes, `use` statements in traits should be avoided, since they can clash with the ones defined in the class that uses the trait. Using a trait should be transparent to the client class, and the only way to do it is to use FQCN in traits.

Closing as @hacfi asks

hacfi:
Good to know..thanks for the explanation @alch

PR #618, opened 2015-03-16
Wizard request shipping range

rogergros:
The wizard setup not asks to fill the shipping range and assumes that a default carrier is set

Ping @elcodi/owners


src/Elcodi/Plugin/StoreSetupWizardBundle/Controller/WizardController.php
```diff
@@ -44,6 +46,17 @@ public function viewAction()
             ->get('elcodi_templates.wizard_status.service');
         $stepsFinished       = $wizardStatusService->getStepsFinishStatus();
 
+        $firstCarrier = $this
+            ->get('elcodi.repository.carrier')
+            ->findOneBy(
+                ['enabled' => true],
+                ['id' => 'ASC']
```
mmoreram: ASC? Mi no understand anything :dancer:

rogergros: I want to get one carrier and I try to get always the oldest one. The ASC is to get the results ordered and taking the first one :)

```php
/**
 * Finds a single entity by a set of criteria.
 *
 * @param array $criteria
 * @param array|null $orderBy
 *
 * @return object|null The entity instance or NULL if the entity can not be found.
 */
public function findOneBy(array $criteria, array $orderBy = null)
{
    // ...
}
```

mmoreram: Hmmm, understood.

PR #622, opened 2015-03-17
Obtain source currency outside the for

arkaitzgarro:
Source currency is always the same inside the for, so it doesn't make sense obtain it in each iteration.


hacfi:
:+1:

Issue #623, opened 2015-03-18
[Media] Invalid image mime type


seventhseeker:
Sometimes valid images have the mime type "application/octet-stream".

"ImageManager" does not accept this mime type, and throws an "InvalidImageException":
https://github.com/elcodi/elcodi/blob/master/src/Elcodi/Component/Media/Services/ImageManager.php#L95

Sample image:
![applicationoctetstreamimage](https://cloud.githubusercontent.com/assets/1076699/6706237/f4f5df94-cd5b-11e4-9e84-28ab432dad2b.jpg)

javiereguiluz:
The problem is that `application/octet-stream` can be anything, not only images. So you cannot blindly trust this MIME type as valid for images.

In theory, the `image/*` constraint is the only valid way to limit image MIME types (this is the one used by Symfony por example). A possible solution: when receiving an `application/octet-stream` "thing", save it in the filesystem and use PHP functions to get information about the file and verify if it's an image or not.

mrzard:
This should be fixed by PR#624

rogergros:
Hi @seventhseeker , could you confirm that's already fixed?

mmoreram:
ping @seventhseeker

seventhseeker:
Hi @rogergros, @mmoreram, sorry for the delay.
Awesome! it works great, thanks very much @mrzard

rogergros:
Great :) closing!

PR #624, opened 2015-03-18
Issue 623: Try to correctly manage images with MIME application/octet-stream

mrzard:
Hi guys, this tries to solve issue #623 

Basically, in `ImageManager`, if MIME `application/octet-stream`
is detected, `getimagesize` will be ran on the file in an attempt to
extract is 'real' MIME type.

I am open to modify this, as this does a couple of 'ugly' things:
- Tests write to sys tmp dir using `file_put_contents`
- Ugly `!isset` to check if we ran `getimagesize` already

Keep up the good work!

### DIFF
--- a/src/Elcodi/Component/Media/Services/ImageManager.php
+++ b/src/Elcodi/Component/Media/Services/ImageManager.php
@@ -92,6 +92,11 @@ class ImageManager
     {
         $fileMime = $file->getMimeType();
 
+        if ('application/octet-stream' === $fileMime) {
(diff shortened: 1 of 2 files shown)


alch:
:+1: Nice one! As regards the `!isset`, once PHP will get the [null coalesce operator](https://wiki.php.net/rfc/isset_ternary) we'll restyle some code

mmoreram:
@mrzard Great dude! Thanks :dancer:

mrzard:
Happy to contribute!

PR #632, opened 2015-03-23
Fixed outbound crop resize

rogergros:
Fixed outbound crop resize.

Applied solution found on http://stackoverflow.com/questions/18762252/how-to-scale-and-crop-with-imagemagick-convert


rogergros:
Fixes https://github.com/elcodi/bamboo/issues/247

PR #635, opened 2015-03-24
Feature/cs fixer config

rogergros:
Cs fixer configuration proposal:
- Applied to `"src/"` folder
- Symfony level [(More info)](http://cs.sensiolabs.org/#usage)
- Extra fixers
  - Align double arrow symbols in consecutive lines.
  - Align equals symbols in consecutive lines.
  - Concatenation should be used with at least one whitespace around
  - Multi-line whitespace before closing semicolon are prohibited.
  - PHP array's should use the PHP 5.4 short-syntax.

@elcodi/owners I've also applied the fixers to the code, let me know if you think that we should apply a different config ;)

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,16 @@
+<?php
+
+return Symfony\CS\Config\Config::create()
+    // use SYMFONY_LEVEL:
+    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
+    // and extra fixers:
+    ->fixers(array(
+        'concat_with_spaces',
+        'multiline_spaces_before_semicolon',
+        'short_array_syntax'
+    ))
+    ->finder(
+        Symfony\CS\Finder\DefaultFinder::create()
+            ->in('src/')
+    )
+;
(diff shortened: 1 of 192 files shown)


javiereguiluz:
Great change and a very intelligent move! Any modern PHP project that doesn't use PSR-2 and/or Symfony CS looks weird. Even Laravel is considering moving to PSR-2 very soon (probably for 5.1 version).

mmoreram:
@javiereguiluz We were already complaining with PSR-2, right?

mmoreram:
BTW, I don't like the spaces to make everything aligned.

rogergros:
@mmoreram In my opinion both "Align double arrow symbols in consecutive lines." and "Align equals symbols in consecutive lines." improve readability but it's something we can discuss.
About the other extra fixers, do you agree?

javiereguiluz:
@mmoreram you are right about aligning spaces. I still like them, but Symfony recently made the switch and now nothing is artificially aligned with white spaces.

xphere:
- :+1: for specific `.php_cs`, we need everybody to work with the same setup, and that's hard without this.
- :-1: for alignment both in assignment and arrows, I prefer smaller changelogs over the small readability this can add. Not for me, thought.
- :+1: for spaces around concatenation, sometimes I miss this operation because it's hidden between its operands.
- :+1: for removing trailing spaces, no need for that, just the extra line at EOF.
- :+1: for 5.4 arrays, this improves readability by a huge amount, and should be in.

rogergros:
Ok then, I also prefer the spaces but most of you think that's not a good idea and it seems that the symfony community is on the same direction so I'll remove all the alignments fixes ;)

rogergros:
Changed :)

Issue #581, opened 2015-03-02
[Product] Logical error with ElcodiProductStock


hacfi:
I think I found a logical error:
https://github.com/elcodi/elcodi/blob/master/src/Elcodi/Component/Product/Services/ProductCollectionProvider.php#L126-144
`$infiniteStockIsNull = is_null(ElcodiProductStock::INFINITE_STOCK);` will always be true. It doesn’t make sense to check a constant for a value. I double-checked and this value also never changed so either $infiniteStockIsNull becomes a method argument or remove it at all!?

mmoreram:
Lol, absolutely ... Thanks for the report :)

hacfi:
You’re welcome! I’m getting more involved again soon!

hacfi:
Closing as #636 resolved this

PR #636, opened 2015-03-24
Fixed error with infinite stock getting offers

rogergros:
Proposed solution for https://github.com/elcodi/elcodi/issues/581

Ping @elcodi/owners

### DIFF
--- a/src/Elcodi/Bundle/ProductBundle/Resources/config/services.yml
+++ b/src/Elcodi/Bundle/ProductBundle/Resources/config/services.yml
@@ -12,4 +12,4 @@ services:
         class: %elcodi.provider.product_collection.class%
         arguments:
             product_repository: @elcodi.repository.product
-            configuration_manager: @elcodi.manager.configuration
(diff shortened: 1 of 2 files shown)


src/Elcodi/Bundle/ProductBundle/Resources/config/services.yml
```diff
@@ -12,3 +12,4 @@ services:
         class: %elcodi.provider.product_collection.class%
         arguments:
             product_repository: @elcodi.repository.product
+            configuration_manager: @elcodi.manager.configuration
```
xphere: I'd rather inject directly the configuration value with `@=elcodi_config('product.use_stock')`

mmoreram: Already fixed in my work.

mmoreram: Ups, nopes.

Please, inject the value instead of all Manager :)

rogergros:
Fixed :)

PR #641, opened 2015-03-26
can unselect purchasable

jromero86:


mmoreram:
ping @elcodi/deciders 

Any comment on that?

alch:
@jromero86 AFAIK in your implementation you extended `Coupon` to have a `oneToMany` relation with a `PurchasableInterface` (so that you can relate a `Coupon` with a `Product` or a `Variant`).

Can you explain your use case to everyone?

jromero86:
In my case, a coupon can be applied directly to a product or a variant. But it is not mandatory if is a general discount.

Right now, if you select a purchasable then you cannot deselect because the variable that is sent to setPurchasable must be of type PurchasableInterface interface and we send a null value.

src/Elcodi/Component/Cart/Entity/Traits/PurchasableWrapperTrait.php
```diff
@@ -55,8 +55,12 @@
      *
      * @return $this Self object
      */
-    public function setPurchasable(PurchasableInterface $purchasable)
+    public function setPurchasable(PurchasableInterface $purchasable = null)
     {
+        if (!$purchasable) {
+            $this->setVariant();
```
xphere: I'd rather make the `null` explicit: `$this->setVariant(null)`

mmoreram: +1

alch:
@jromero86 as we discussed, I don't think it is appropriate to use the `PurchasableWrapperTrait` only as a shortcut to have the `setProduct` and `setVariant` methods for free: this trait and the corresponding interface is used **only** for *things that can be bought*.
In your implementation you are using the trait with the `Coupon` entity, but the `Coupon` is not meant to be *added to the Cart*.
In this case it's better to add the methods in your overridden Coupon class.

Issue #292, opened 2014-10-10
[RFC] Discuss the existence of NullMoney


mmoreram:
* And how this can co-exists with the Symfony Form Component

xphere:
Haven't we decided on the removal of `NullMoney` yet?

alch:
+1 for removal. I think that the best thing to do is to open a PR and discuss there

cordoval:
bump @mmoreram close?

PR #653, opened 2015-03-27
Removed NullMoney and StubMoney

mmoreram:
* In order to simplify the model


mmoreram:
Solves https://github.com/elcodi/elcodi/pull/653

Related to https://github.com/elcodi/elcodi/issues/292 as well

mmoreram:
ping @elcodi/deciders

xphere:
I think that `Money` should be rethought as a whole. But not for v1.0, so I'm :+1:

alch:
Agree 100%

@xphere why do you think it should be rewritten'

xphere:
I think it adds too many complexity for what it does, wrapping another `Money` library under the hood, moving back and forth between the two codebases. IMO, there has to be a simple way.

PR #655, opened 2015-03-27
Removed ReferralProgram Bundle and Component

mmoreram:
* Because both are not complying desired code standards and need some love and refactoring, these
packages are removed with the desire of adding them back as plugin


mmoreram:
ping @elcodi/deciders

xphere:
I'm :+1: on this since the first minute I joined the Elcodi team. :grin:
cof cof plugin cof cof

alch:
Ping @claretcrab

claretcrab:
:+1: we already talk with Marc and it's ok, if we need this feature we will implement it as a plugin for deliberry based on this removed code

PR #661, opened 2015-03-27
Added v0.5.x changelog

mmoreram:


xphere:
Some improvements in format.

- Hashes should go between `` to make them monospace and take the same width.
- Maybe there should be a link to the actual commit?
- Authorship is important, but I think this adds too much noise here, solved by linking to the commit.
- Non-core contributors can be listen at the end as a list with "Thanks to @​Bob, @​Joe and @​Jane for contributing to vx.y.z"

rogergros:
Nice work :+1:

claretcrab:
great!

javiereguiluz:
@mmoreram thanks for your work creating this great and detailed changelog information. I really like it!

mmoreram:
:dancer:

PR #662, opened 2015-03-29
Disable default registerCommands method

hacfi:
Just a small improvement to prevent https://github.com/symfony/symfony/blob/2.6/src/Symfony/Component/HttpKernel/Bundle/Bundle.php#L173-197 from being executed on every bundle although we register all commands as services.

### DIFF
--- a/src/Elcodi/Bundle/CartBundle/ElcodiCartBundle.php
+++ b/src/Elcodi/Bundle/CartBundle/ElcodiCartBundle.php
@@ -17,13 +17,13 @@
 
 namespace Elcodi\Bundle\CartBundle;
 
+use Symfony\Component\Console\Application;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
 use Elcodi\Bundle\CartBundle\CompilerPass\MappingCompilerPass;
 use Elcodi\Bundle\CartBundle\DependencyInjection\ElcodiCartExtension;
-use Elcodi\Bundle\CoreBundle\ElcodiCoreBundle;
 use Elcodi\Bundle\CoreBundle\Interfaces\DependentBundleInterface;
 
 /**
@@ -67,5 +67,17 @@ class ElcodiCartBundle extends Bundle implements DependentBundleInterface
             '\Elcodi\Bundle\ConfigurationBundle\ElcodiConfigurationBundle',
             '\Elcodi\Bundle\CoreBundle\ElcodiCoreBundle',
         ];
+    }
+
+    /**
+     * Register Commands.
+     *
+     * Disabled as commands are registered as services.
+     *
+     * @param Application $application An Application instance
+     */
+    public function registerCommands(Application $application)
+    {
+        return;
     }
 }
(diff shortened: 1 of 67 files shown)


src/Elcodi/Bundle/CartBundle/ElcodiCartBundle.php
```diff
@@ -17,13 +17,13 @@
...
 namespace Elcodi\Bundle\CartBundle;
 
+use Symfony\Component\Console\Application;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
 use Elcodi\Bundle\CartBundle\CompilerPass\MappingCompilerPass;
 use Elcodi\Bundle\CartBundle\DependencyInjection\ElcodiCartExtension;
-use Elcodi\Bundle\CoreBundle\ElcodiCoreBundle;
```
hacfi: Didn’t want to create a separate PR just for this one unused class

mmoreram: Never mind :)

mmoreram:
Hmm, well, this PR is incompleted. If you add these changes, then you must add `symfony/console` as a dependency in all bundles ;)

hacfi:
@mmoreram Why? Have a look at the original bundle code at https://github.com/symfony/symfony/blob/2.6/src/Symfony/Component/HttpKernel/Bundle/Bundle.php#L173-197 . This PR even loosens the dependency.

hacfi:
Maybe Noback’s article at http://php-and-symfony.matthiasnoback.nl/2013/10/symfony2-console-commands-as-services-why/ helps to explain the reason behind this PR.

xphere:
Maybe this even fixes the problem with `Command` dependencies having to be instanciated?

mmoreram:
Yes, I understand the why of this PR, and makes sense... But take a look here

https://github.com/symfony/symfony/blob/2.6/src/Symfony/Component/HttpKernel/composer.json#L30

If you use a namespace of a package, then you are coupled to this package. If we want this change... then you need dependency with this package.

mmoreram:
Yes

hacfi:
@mmoreram True! But aren’t you depending on it already? https://github.com/hacfi/elcodi/blob/master/src/Elcodi/Bundle/AttributeBundle/ElcodiAttributeBundle.php#L30 is already extending `Symfony\Component\HttpKernel\Bundle\Bundle` so I’m not adding a class which hasn’t been used before.

mmoreram:
No, you are depending on HttpKernel, not Console ;) You should not depend on what HttpKernel depends ;)

hacfi:
@mmoreram Let me check if I really changed a dependency here. As far as I understand that class extends `Symfony\Component\HttpKernel\Bundle\Bundle` which depends on that same Console class: https://github.com/symfony/symfony/blob/2.6/src/Symfony/Component/HttpKernel/Bundle/Bundle.php#L17
Will get back to you tonight!

mmoreram:
Yes, but you cannot depend on that dependency. Your dependencies will always be all resources you need, even if you know that your dependencies have X dependencies.

mmoreram:
What I would do is to add the dependency in all Bundles and override this method :)

hacfi:
@mmoreram Great...will update the PR now.

hacfi:
While I was at it I fixed some inconsistencies with the sorting and removed a dependency which was listed twice.

mmoreram:
@hacfi Great :dancer:

PR #671, opened 2015-03-30
Sebastian/Money update

fjavier-moreno:
Add, among others, getConvertedAmount method.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
         "doctrine/doctrine-cache-bundle": "~1.0",
         "mmoreram/simple-doctrine-mapping": "~0.1",
 
-        "sebastian/money": "~1.3",
+        "sebastian/money": "1.5.*",
         "knplabs/knp-gaufrette-bundle": "~0.1",
         "ircmaxell/password-compat": "~1.0",
         "guzzlehttp/guzzle": "~5.2",


composer.json
```diff
@@ -60,7 +60,7 @@
         "doctrine/doctrine-cache-bundle": "~1.0",
         "mmoreram/simple-doctrine-mapping": "~0.1",
 
-        "sebastian/money": "~1.3",
+        "sebastian/money": "1.5.*",
```
mmoreram: Does Sebastian's Money implementation promises BC Break?

fjavier-moreno: Sebastian does not comment any incompatibility with older versions.

mmoreram: I am talking about BC Break Promise. That means that you can use `~1.5` with any fear of BC Break.

fjavier-moreno: I'm talking about this update only add new methods. Not change any previous methods used by Elcodi and add some news like getConvertedAmount, that is very useful when you work with API's and Twig extension is not valid to format php output.

mmoreram: I know. Just saying that if you put `1.5.*` then you cannot use `1.6.*`, and this is not the behavior we want.

If you want this change to be merged, please use the `~` symbol -> `~1.5`

alch:
@fjavier-moreno @mmoreram I don't think there will be a breaking change for a given minor version `1.5.*`
Anyway why not stick with the `~1.3` version and then accept the PR?

xphere:
If we rely on `getConvertedAmount` for anything, we need to upgrade dependency to `~1.5`.
If this is project-specific, then there's no need for that in elcodi nor bamboo.

mmoreram:
@fjavier-moreno why did you close this PR?

jromero86:
You how to calculate discounts if given a number with more than two decimal then?

PR #680, opened 2015-04-01
[Configuration] Integrate with ControllerExtraBundle

xphere:
Add a new `Configuration` annotation for `mmoreram/controller-extra-bundle`. If you don't have this bundle, does nothing.

```php
/**
 * @Configuration(
 *     name = "parameter",
 *     key = "key_from_configuration",
 *     default = "default_value"
 * )
 */
public function testAction($parameter)
{
    …
}
```

This extracts `key_from_configuration` current value from `Configuration` and sets `$parameter` with it, defaulting to `default_value` if it does not exist. In case no default value is supplied and configuration can't be found, throws an exception.

### DIFF
--- a/src/Elcodi/Bundle/ConfigurationBundle/Annotation/Configuration.php
+++ b/src/Elcodi/Bundle/ConfigurationBundle/Annotation/Configuration.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of the Elcodi package.
+ *
+ * Copyright (c) 2014 Elcodi.com
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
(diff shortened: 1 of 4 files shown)


xphere:
Will it make sense to have `src/Elcodi/Bundle/ConfigurationBundle/Annotation/Configuration.php` and `src/Elcodi/Bundle/ConfigurationBundle/Resolver/ConfigurationResolver.php` in the Configuration component and the rest in the bundle?

I think it's ok to have all of them in the bundle, but maybe I'm wrong.

mmoreram:
@xphere What's the final strategy? Are you moving this to Bamboo common namespace?

xphere:
Yep, sorree. I'm closing this in favor of moving everything to Bamboo, as discussed yesterday.

mmoreram:
Great dude :dancer:

PR #695, opened 2015-04-04
Removed message from Exception assertion

mmoreram:
* Otherwise, tests fail


mmoreram:
@rogergros A test should be as light as possible. It means that it should not be affected with an exception message tests, for example (in this case)

cordoval:
you mean britle

PR #697, opened 2015-04-04
Improved AddressManager class

mmoreram:
- Main method returns saved address, instead of having `returnLastWhatever` method


mmoreram:
ping @rogergros 

I don't really understand why this method...

PR #699, opened 2015-04-05
[WIP] make name translatable just means removing from core the mapping on name field on Rule

cordoval:
|Q            |A   |
|---          |--- |
|Bug Fix?     |no  |
|New Feature? |no  |
|BC Breaks?   |no  |
|Deprecations?|no  |
|Tests Pass?  |no  |
|Fixed Tickets|#643|
|License      |MIT |

### DIFF
--- a/src/Elcodi/Component/Rule/Repository/RuleRepository.php
+++ b/src/Elcodi/Component/Rule/Repository/RuleRepository.php
@@ -27,14 +27,14 @@ use Elcodi\Component\Rule\Entity\Interfaces\RuleInterface;
 class RuleRepository extends EntityRepository
 {
     /**
-     * Return a rule by name
+     * Return a rule by id
      *
-     * @param string $name
+     * @param string $id
      *
      * @return RuleInterface|null
      */
-    public function findOneByName($name)
+    public function findOneById($id)
     {
-        return parent::findOneBy(['name' => $name]);
+        return parent::findOneBy(['id' => $id]);
     }
 }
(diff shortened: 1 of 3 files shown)


cordoval:
@mmoreram  removing the mapping for name to make it translatable creates problems. I think i say it is better to create a new property label and then leave name alone. We can then translate label and such but we keep a rule name in english or whatever short name in any language in which the domain expert is staying for development fixed.

cordoval:
@mmoreram could you please respond to this question? That will prevent me to go in the wrong approach, right now i think i should just add a new property named label and since it does not have mapping then this will work nicely.

xphere:
The `name` field can't be translated, since this is used internally by rule validation in `ExpressionLanguage` and will cause problems when changing languages. We need `name` to be a identifier-like string (maybe a dashed separated, lowercased string) and a separated `label` field which can be safely translated.

cordoval:
correct, that is what i ran into. I believe name should be left intact as it is now. And then as you and i agreed we need a ```label``` field that can be translated. So i will adapt PR accordingly. :+1:

src/Elcodi/Component/Rule/Repository/RuleRepository.php
```diff
@@ -27,14 +27,14 @@
...
-     * @param string $name
+     * @param string $id
      *
      * @return RuleInterface|null
      */
-    public function findOneByName($name)
+    public function findOneById($id)
     {
-        return parent::findOneBy(['name' => $name]);
+        return parent::findOneBy(['id' => $id]);
```
mmoreram: I don't really understand this method. Without this implementation you can call to `findOneById` as well, and the result will be the same. In fact, you can call `find` and will return the same result...

PR #700, opened 2015-04-05
remove randomized based url sharding for assets (images) fix #621

### LINKED ISSUE #621: ImageExtension::randomizeBaseUrls not so random
The distribution of current `randomizeBaseUrls` in `ImageExtension` is quite poor and scales horribly, check [this test](http://3v4l.org/FLaC5) with [current implementation](https://github.com/elcodi/elcodi/blob/ed36afb74780c2d4d1b80f496d2dc20dea7cd9f5/src/Elcodi/Component/Media/Twig/ImageExtension.php#L184).

There are some issues:
- `0` is never picked.
- `1` and `2` are more likely to be chosen (like 10:1 and 3:1 respectively)
- Multiple calls for same `path`, can return different shard.

I'd remove the `substr` call and do just `mt_rand(0, count($hostNames))`

cordoval:
|Q            |A  |
|---          |---|
|Bug Fix?     |no |
|New Feature? |no |
|BC Breaks?   |no |
|Deprecations?|no |
|Tests Pass?  |no |
|Fixed Tickets|  #621  |
|License      |MIT|

### DIFF
--- a/src/Elcodi/Bundle/MediaBundle/Resources/config/twig.yml
+++ b/src/Elcodi/Bundle/MediaBundle/Resources/config/twig.yml
@@ -9,7 +9,5 @@ services:
             router: @router
             image_resize_controller_route_name: %elcodi.core.media.image_resize_controller_route_name%
             image_view_controller_route_name: %elcodi.core.media.image_view_controller_route_name%
-            images_domain_sharding_enabled: %elcodi.core.media.images.domain_sharding.enabled%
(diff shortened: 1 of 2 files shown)


cordoval:
do you know what is wrong with github?
https://github.com/elcodi/elcodi/pull/699
has a second commit that does not show on the PR
https://github.com/cordoval/elcodi/commits/643-rule-add-translatable-in-rule-field-name

@mmoreram github is messed up

Issue #521, opened 2015-02-17
[Media] Create image on resize


sfblaauw:
Wil be :+1:

xphere:
Can you elaborate on this one, please?

rogergros:
@sfblaauw ? :)

alch:
This is a complex topic, we should open an RFC. The first problem to be solved is the following (we assume that no reverse proxy for image caching is used, which is **not recomended**)

1) The application receives a request for `/images/21/resize/291/372/5.jpg`
2) The `image` with id 21 is fetched from doctrine and gaufrette is used to retrieve the associated image
3) `Imagemanager::resize()` is used to resize the image using the appropriate adapter (currently `ImageMagick`)
4) Symfony returns the resized image with the correct `content-type` and `Etag` and `Expires` / `Cache-Control: max-age` headers

Where can we put the logic that stores the resized file? An option can be to do it between step `3` and `4`. However we need to save this information (that the image has been correctly resized and saved at this resolution, with the correct type, like `ElcodiMediaImageResizeTypes::FORCE_MEASURES` for instance)

This implies that every `Image` entity had a `oneToMany` relation with a new entity, say `ResizedImage`, with at least the following fields:

```
resized_image_id
image_id
width
height
resize_type
```
Should we have this information, it is possible to reshape the image resize process to:

1) The application receives a request for `/images/21/resize/291/372/5.jpg`
2) The `image` with id 21 is fetched from doctrine and gaufrette is used to retrieve the associated image
3) `ImageManager` checks whether a `ResizedImage` exists for this `Image` with the matching resize parameters.
4) If there is such entity, it means that the `Image` had already be stored with these parameters. Gaufrette will then retrieve the `resized_image_id` and the `ImageManager` can return the image with no further processing.
5) If no such entity exists, proceed with using `ImageManager::resize()` to resize the image using the appropriate adapter (currently `ImageMagick`)
6) Create a new `ResizedImage`, persist it, add it to the `Image` collection, flush and use gaufrette to physically store the resized image with the new `resized_image_id`
7) Symfony returns the resized image with the correct `content-type` and `Etag` and `Expires` / `Cache-Control: max-age` headers

This is far from being perfect but it will allow cache misses to have a far lesser impact on CPU and disk

mmoreram:
I will add this issue for next `v1` release. Good explanation @alch.

cordoval:
[...]

PR #701, opened 2015-04-05
[WIP] media create image on resize

cordoval:
|Q            |A   |
|---          |--- |
|Bug Fix?     |no  |
|New Feature? |no  |
|BC Breaks?   |no  |
|Deprecations?|no  |
|Tests Pass?  |no  |
|Fixed Tickets|#521|
|License      |MIT |

### DIFF
--- a/src/Elcodi/Bundle/MediaBundle/Resources/config/doctrine/ResizedImage.orm.yml
+++ b/src/Elcodi/Bundle/MediaBundle/Resources/config/doctrine/ResizedImage.orm.yml
@@ -1,0 +1,37 @@
+Elcodi\Component\Media\Entity\ResizedImage:
+    type: entity
+    repositoryClass: Elcodi\Component\Media\Repository\ImageRepository
+    table: image
+    id:
+        id:
+            type: integer
+            generator:
+                strategy: AUTO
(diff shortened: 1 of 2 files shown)


mmoreram:
@cordoval Do you think is necessary a new entity for that?

About your question, we're using collections in entities. What we don't want to couple is entities and any collection implementation, like ArrayCollection.

xphere:
This is related to a discussion we're having in #521, but we have not yet reached an agreement.

cordoval:
@xphere in FOSS we use the table PR as in this PR description to display the ticket that is related to. Maybe you are using a client for github?  but if you are on this page to see the discussion this is already in the description of the PR so right. And correct we haven't reached an agreement, just exploring the idea publicly to see what roadblocks we encounter.

xphere:
Yep, just pointing to the discussion in another thread, hoping for some additional visibility.

cordoval:
:+1:

mmoreram:
I don't think that a new table for such feature is a good idea. In such table won't be relevant information indeed. Maybe I don't understand properly the implementation, but I think that, as said in #521, a new service overwriting the manager would be enough.

PR #702, opened 2015-04-05
create only one command for all sitemaps

cordoval:
|Q            |A   |
|---          |--- |
|Fixed Tickets|#468|

PR #712, opened 2015-04-09
Feature/sitemap refactor

mmoreram:


mmoreram:
ping @cordoval This PR replaces #702 (With your commits as well)
ping @elcodi/owners Refactor of this component

src/Elcodi/Bundle/SitemapBundle/DependencyInjection/ElcodiSitemapExtension.php
```diff
@@ -92,10 +90,9 @@ protected function postLoad(array $config, ContainerBuilder $container)
     {
         $this
             ->loadBlocks($config, $container)
```
xphere: If each one works with a specific key, I'd suggest to pass only that, like:
```php
$this
    ->loadBlocks($config['blocks'], $container)
    ->loadStatics($config['statics'], $container)
    ->loadBuilders($config['builders'], $container)
    ->loadProfiles($config['profiles'], $container)
;
```

src/Elcodi/Bundle/SitemapBundle/DependencyInjection/ElcodiSitemapExtension.php
```diff
@@ -113,13 +110,16 @@ protected function loadBlocks(array $config, ContainerBuilder $container)
         foreach ($blocks as $blockName => $block) {
             $container
                 ->register(
-                    'elcodi.sitemap_entity_loader.' . $blockName,
-                    '%elcodi.core.sitemap.loader.entity_loader.class%'
+                    'elcodi.sitemap_element_generator.entity_' . $blockName,
+                    'Elcodi\Component\Sitemap\Element\EntitySitemapElementGenerator'
                 )
+                ->addArgument(new Reference('elcodi.factory.sitemap_element'))
```
xphere: I'd group all of this in a single `setArguments` call.
Maybe is easier to create a new `Definition`, which allows for arguments in its constructor, and then call `$container->addDefinition($definition)`?

src/Elcodi/Component/Sitemap/Builder/SitemapBuilder.php
```diff
@@ -0,0 +1,126 @@
...
+     * @param string|null $language Language
+     *
+     * @return string Generated data
+     */
+    public function build($language = null)
+    {
+        $sitemapElements = [];
+
+        /**
+         * @var SitemapElementGeneratorInterface $sitemapElementGenerator
```
xphere: With [this typehint](https://github.com/elcodi/elcodi/pull/712/files#diff-ad3ac0922fdfce608f813798d73fc47cR29) this one should not be needed.

src/Elcodi/Component/Sitemap/Builder/SitemapBuilder.php
```diff
@@ -0,0 +1,126 @@
...
+                $sitemapElements,
+                $sitemapElementGenerator->generateElements($language)
+            );
+        }
+
+        $data = $this
+            ->sitemapRenderer
+            ->render($sitemapElements);
+
+        file_put_contents(
```
xphere: This is hardly testable and now it's impossible to generate a sitemap without dumping to filesystem.
I'd go with a `SitemapDumper`, maybe decorating `SitemapBuilder` (which may need the creation of a `SitemapBuilderInterface`)

src/Elcodi/Component/Sitemap/Command/SitemapBuildCommand.php
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * @return integer|null|void
+     *
+     * @throws SitemapBuilderNotFoundException Profile not found
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $builderName = $input->getArgument('builder-name');
+        $language = $input->getOption('language');
+        $sitemapBuilder = $this
```
xphere: This will throw a `ServiceNotFoundException` when the sitemap does not exist.

src/Elcodi/Component/Sitemap/Command/SitemapProfileCommand.php
```diff
@@ -0,0 +1,94 @@
...
+     * @param OutputInterface $output The output interface
+     *
+     * @return integer|null|void
+     *
+     * @throws SitemapProfileNotFoundException Profile not found
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $profileName = $input->getArgument('profile-name');
+        $sitemapProfile = $this
```
xphere: Same as [this](https://github.com/elcodi/elcodi/pull/712/files#r28046911)

src/Elcodi/Component/Sitemap/Element/EntitySitemapElementGenerator.php
```diff
@@ -0,0 +1,141 @@
...
+     * @param ObjectRepository            $repository            Repository
+     * @param string                      $method                Method
+     * @param array                       $arguments             Arguments
+     * @param string|null                 $changeFrequency       Change frequency
+     * @param string|null                 $priority              Priority
+     */
+    public function __construct(
+        SitemapElementFactory $sitemapElementFactory,
+        SitemapTransformerInterface $transformer,
+        ObjectRepository $repository,
```
xphere: `repository`, `method` and `arguments` can be collapsed into a single `callable` parameter

src/Elcodi/Component/Sitemap/Profile/SitemapProfile.php
```diff
@@ -17,88 +17,68 @@
...
      */
-    protected $entityLoaders;
+    protected $languages;
 
     /**
-     * @var string
+     * @var SitemapBuilder[]
      *
-     * Path
+     * Array of SitemapBuilders
```
xphere: Maybe it's just me, but I found this kind of comments exist only to "pass" the PHPDoc requirements. For me this stands at the same level of usefullness as `++$i; // increments i`. And it's not only this one: we're spreading pointless comments like this all over the place.

src/Elcodi/Bundle/SitemapBundle/DependencyInjection/ElcodiSitemapExtension.php
mmoreram: Agree

src/Elcodi/Bundle/SitemapBundle/DependencyInjection/ElcodiSitemapExtension.php
mmoreram: I don't understand... :(

src/Elcodi/Component/Sitemap/Builder/SitemapBuilder.php
mmoreram: Indeed it is already like you say :) I remove this line. Thanks.

src/Elcodi/Component/Sitemap/Command/SitemapBuildCommand.php
mmoreram: Maybe the service exists but is not a SitemapBuilder

src/Elcodi/Component/Sitemap/Command/SitemapProfileCommand.php
mmoreram: Sames as [this](https://github.com/elcodi/elcodi/pull/712/files#r28051154)

src/Elcodi/Component/Sitemap/Element/EntitySitemapElementGenerator.php
mmoreram: So... who exactly creates the callable code? The extension? Another service?

src/Elcodi/Component/Sitemap/Builder/SitemapBuilder.php
mmoreram: Well, a Dumper decorating a Builder and implementing a Builder Interface? Maybe is better not to decorate but inject.

src/Elcodi/Component/Sitemap/Command/SitemapBuildCommand.php
xphere: This doesn't look consistent to me. I mean, we throw a `ServiceNotFoundException` if the service does not exist, but a `SitemapBuilderNotFoundException` if the service exists but it's not a `SitemapBuilder`?

I'd extract the service acquisition to a protected method helper, just like:
```php
public function getSitemapBuilderFor($name)
{
    $serviceId = "elcodi.sitemap_builder.{$builderName}";
    $service = $this->container->has($serviceId) ? $this->container->get($serviceId) : null;
    if ($service instanceof SitemapBuilder) {
        return $service;
    }
    throw new SitemapBuilderNotFoundException(sprintf(
        'Sitemap builder "%s" not found',
        $builderName
    ));
}
```

src/Elcodi/Bundle/SitemapBundle/DependencyInjection/ElcodiSitemapExtension.php
xphere: Instead of many calls to `addArgument`, make it just one to `setArguments`.
Or even better (IMO) to create a `Definition` and call `setDefinition` directly.

src/Elcodi/Component/Sitemap/Command/SitemapBuildCommand.php
mmoreram: It is not really important @xphere. I will remove the exceptions and that's all.

src/Elcodi/Bundle/SitemapBundle/DependencyInjection/ElcodiSitemapExtension.php
mmoreram: I don't understand the second one. Can you add an example please?

src/Elcodi/Component/Sitemap/Builder/SitemapBuilder.php
xphere: Please note the use of the word "maybe" :stuck_out_tongue:

I find the `SitemapBuilder` class having too many responsabilities:
- Collect elements from generators
- Render this elements to a specific format
- Dump the rendered content to a file

Maybe it's that I don't understand the `Builder` prefix.

mmoreram: Yes, you're right. I am already fixing it. Thanks.

src/Elcodi/Bundle/SitemapBundle/DependencyInjection/ElcodiSitemapExtension.php
xphere: This one would be:
```php
$definition = new Definition(
    'Elcodi\Component\Sitemap\Element\EntitySitemapElementGenerator',
    [
        new Reference('elcodi.factory.sitemap_element'),
        new Reference($block['transformer']),
        new Reference($block['repository_service']),
        $block['method'],
        $block['arguments'],
        $block['changeFrequency'],
        $block['priority'],
    ]
);
$definition->setPublic(false);
$container->setDefinition(
    "elcodi.sitemap_element_generator.entity_{$blockName}",
    $definition
);
```

src/Elcodi/Component/Sitemap/Command/SitemapBuildCommand.php
xphere: :+1:

src/Elcodi/Component/Sitemap/Element/EntitySitemapElementGenerator.php
xphere: The thing is to decouple the `Generator` from `ObjectRepository`.
The problem is the `Extension` can't add closures to the `Container` (as far as I know :cry:)

mmoreram: I understand... I thought about it yesterday, but there is a way :)

src/Elcodi/Component/Sitemap/Profile/SitemapProfile.php
mmoreram: Yes, they are. And like this one, a lot, but take in account PHPDocumentor and others ;)

src/Elcodi/Bundle/SitemapBundle/DependencyInjection/ElcodiSitemapExtension.php
mmoreram: Take in account that this code is computed in cache-warm-up time, so it is not really important at all... Both ways are right indeed, so if you want to purpose a full change of this strategy (with pros and contras), then you can expose your POV in a RFC issue :)

mmoreram:
ping @elcodi/deciders Changes done.

alch:
Looks good :+1: 

The only thing that I would discuss (not now) is to avoid the use of `file_get_contents` or anyway to rely on a concrete filesystem that is not ubuquitous (like S3 or Google Cloud Storage), because PaaS like Heroku or AppEngine have a very limited write support.

Let's start the **anti file_set_contents()** campaign! (like [this](http://antiifcampaign.com/))

PR #718, opened 2015-04-10
Fixed api locations route

rogergros:
Fixed url for locations API

Related with https://github.com/elcodi/bamboo/pull/338


mmoreram:
This is not good at all, because you are using `%bamboo_api_prefix%`, and this is a Bamboo parameter...

cordoval:
can we make it a config param of the bundle that can be defined via the configuration of the bundle?

cordoval:
This is what i mean #721 @mmoreram

rogergros:
Woops, right, the thing is that this is now broken.

We can simply close it, the think is that we have a **%bamboo_api_prefix%** that's used nowhere with this PR https://github.com/elcodi/bamboo/pull/338/files

PR #719, opened 2015-04-10
Added possibility to disable payment methods

rogergros:
- Payment methods can be disabled via config file
- Wizard bundle now depends on payment bridge
- Wizard checks only the payment methods enabled

### DIFF
--- a/src/Elcodi/Plugin/StoreSetupWizardBundle/ElcodiStoreSetupWizardBundle.php
+++ b/src/Elcodi/Plugin/StoreSetupWizardBundle/ElcodiStoreSetupWizardBundle.php
@@ -17,6 +17,7 @@
 
 namespace Elcodi\Plugin\StoreSetupWizardBundle;
 
+use Elcodi\Bundle\CoreBundle\Interfaces\DependentBundleInterface;
 use Symfony\Component\Console\Application;
 use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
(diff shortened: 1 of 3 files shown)


src/Elcodi/Plugin/StoreSetupWizardBundle/ElcodiStoreSetupWizardBundle.php
```diff
@@ -50,4 +51,16 @@ public function registerCommands(Application $application)
...
+
+    /**
+     * Create instance of current bundle, and return dependent bundle namespaces
+     *
+     * @return array Bundle instances
+     */
+    public static function getBundleDependencies()
+    {
+        return [
+            '\Elcodi\Common\PaymentBridgeBundle\PaymentBridgeBundle',
```
cordoval: can we just do without the \?

mmoreram: Yes. @rogergros Please remove the first `\`.

cordoval:
bump @mmoreram resolved in #720

rogergros:
Solved, I've also fixed some aesthetic problems (Fixer).

cordoval:
:+1:

src/Elcodi/Plugin/StoreSetupWizardBundle/Resources/config/services.yml
```diff
@@ -9,6 +9,7 @@ services:
             configuration_manager: @elcodi.manager.configuration
             product_repository: @elcodi.repository.product
             shipping_range_repository: @elcodi.repository.shipping_range
+            enabled_methods: %store.payment_bridge.enabled_methods%
```
mmoreram: Hmmm, mi no like this. Please, rethink this element, as you cannot use any Store parameter

mmoreram: The way of solving this is adding this configuration as a plugin configuration element.

mmoreram: @rogergros

PR #734, opened 2015-04-14
Remove Plugin subtree

xphere:
Remove everything in `src/Elcodi/Plugin`.
We're gonna move this to Bamboo, right where it belongs.


cordoval:
you could have merged my PR first about removing the extra bundle dependency :-1:

mmoreram:
@cordoval Exactly. Now in Bamboo maybe is not necessary... or it is?

cordoval:
i don't know, i would get rid of all those dependencies if you ask me :blush:

PR #750, opened 2015-04-21
Added last-login-date and datetime factory

mmoreram:
* AbstractUser has now lastLoginAt. Is a DateTime, updated with an eventListener
* Created DateTimeFactory as a service. Returns now() in create() method
* Injected DateTimeFactory in AbstractFactory and created now() method
* Changed all `new DateTime()` in all places, in favor of $this->now()
* Fixed some tests

### DIFF
--- a/src/Elcodi/Bundle/AttributeBundle/Resources/config/factories.yml
+++ b/src/Elcodi/Bundle/AttributeBundle/Resources/config/factories.yml
@@ -11,6 +11,7 @@ services:
         class: Elcodi\Component\Attribute\Factory\AttributeFactory
         calls:
             - [setEntityNamespace, ["%elcodi.entity.attribute.class%"]]
+            - [setDateTimeFactory, ["@elcodi.factory.datetime"]]
 
     #
     # Factory for Value entities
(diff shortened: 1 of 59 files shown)


cordoval:
I am not sure if the factory change was really needed here. There are some parts we are perhaps going too far imo. I of course `agree` but would definitely like to see some effort in keeping things simple.

alch:
I am also a little bit puzzled about this. Do we need dependency inversion on generic `Date` objects?

xphere:
It's hard to test classes making `new DateTime`, since it may be inconsistent.

This may fail non-deterministically:
```php
$date = new DateTime();
// Do some more things
$this->assertEqual(new DateTime(), $date);
```

alch:
I don't get it. `DateTimeTrait` has public getters/setters. If the date needs to be injected only for test purposes why not use setters?
Moreover, I would use [parent services](http://symfony.com/doc/current/components/dependency_injection/parentservices.html) to avoid repeating the injection of the datetime factory

xphere:
Yes, with parent services it would be easier to setup, sure. I don't know, maybe @mmoreram did it this way because not every `AbstractFactory` needs the `DateTimeFactory` dependency?

cordoval:
@xphere your example is a bad reason imo. I would use value objects for that and those kind of assertions are done differently in other packages.  I believe the change had a reason but it was not very good imo. Even if you want to move things because of timezone, there could be other better ways to do it.

PR #759, opened 2015-04-24
Refactored Menu engine

mmoreram:
* Using specific event classes with an abstract one
* Using specific Event Dispatcher
* Refactored classes responsibilities
* Removed previous unit test, too coupled
* Added some services tests (definition)


xphere:
Good one! :+1: But I don't like the `MenuManager` to depend on both `serializer` and `processor`, and then `processor` also depending on `serializer`. This cluttered chain is telling us something...

PR #753, opened 2015-04-21
Added back overridable host in image url routes

alch:
PR #700 removed pseudo-randomization of static url (to me more exact, the **host** part of them) for media routes, such as resizing and viewing images.
Although it correctly removed the "sharding" of the hostnames, it also removed *entirely* the possibility to override the host in generated image-resize URLs, which is critical when working with CDNs. For example: mapping `http://www.elcodi.com/images/2536` to `http://cdn.elcodi.com/images/2536`

This PR adds back this feature, and also removes what's left of the *old* pseudo-random configuration

PR #760, opened 2015-04-25
Fixed unwanted change of routing context hostname

alch:
#753 introduced a bug since it modified the routing context without restoring it. This PR fixes it by cloning the routing context.


mmoreram:
Thanks @alch :)

claretcrab:
Great!

claretcrab:
Hi, can we tag a new version please? :D

Issue #716, opened 2015-04-09
Order getCouponAmount fails if couponCurrency is null


claretcrab:
Hi,

Sometimes couponCurrency attribute in order is created as null. This generates an error when you do a getCouponAmount.

The first approach to avoid this is: (Cart Entity)
```
    /**
     * Gets the Coupon amount with tax
     *
     * @return MoneyInterface Coupon amount
     */
    public function getCouponAmount()
    {
        if (isset($this->couponCurrency)) {
            return Money::create(
                $this->couponAmount,
                $this->couponCurrency
            );
        } else {
            return null;
        }
    }
```

But the best approach is to add a cartEventDispatcher before $cart = $this->cartFactory->create(); and give the responsibility to cartCoupon to setCouponAmount as

```
return Money::create(
            0,
            $this->currencyWrapper->getDefaultCurrency()
        );
```

xphere:
Can't we follow the trace to the point where `couponCurrency` should be set and why it's not working?

cordoval:
please @claretcrab clone bamboo and reproduce your problem there and describe the scenario so we can work on it. Thanks

claretcrab:
Hi, non of you can reproduce this behaviour? just open your order tables, i'm sure you will see some of them with null coupon_currency_iso

mmoreram:
ping @claretcrab I will take a look. ASAP we find the reason we must fix the problem and test the scenario. Thanks.

PR #770, opened 2015-05-04
Added coupons to orders when converting from cart

rogergros:
Fixed a bug when converting from cart to order. Coupon amount was not being passed.


claretcrab:
Hi, it's related to https://github.com/elcodi/elcodi/issues/716 ?

mmoreram:
@claretcrab It seems :dancer:

PR #773, opened 2015-05-06
Added respositories methods needed to get subcategories products

rogergros:
- Added a "recursive" option to "getChildrenCategories" methods on the Categories repository
- Added a "getAllFromCategories" method to the Product repository
- Added tests to test both methods


mmoreram:
Great @rogergros :)

PR #774, opened 2015-05-07
Added Metrics Load Command

mmoreram:
* Added loader, repository and command
* Fixed some entities method parameters casting
* Fixed some CS


mmoreram:
ping @rogergros @xphere

PR #777, opened 2015-05-09
Updated service definition

mmoreram:
* Added lazyness to make console engine able to work


cordoval:
does this really help with console service? in symfony all services are loaded up front

mmoreram:
Yes. This service was injected in a console command, and as you can see, this service has an entity manager injected.

Read this - http://mmoreram.com/blog/2015/05/08/lazy-commands-in-symfony/

If the database is not created, then it crashes :( As long as Symfony implementation does not change, we will have to use laziness there :(

cordoval:
yeah i have seen that before.

What i meant is a deeper problem, so this lazyness is just to make this work, but the problem of performance is another one.

mmoreram:
Well, I am not taking care of performance there, as this service, is till now used only in this command.

cordoval:
right

PR #781, opened 2015-05-11
type hinting fix

claretcrab:
There is a bug when a category is created as parent category (parent_id = null)

### DIFF
--- a/src/Elcodi/Component/Product/Repository/CategoryRepository.php
+++ b/src/Elcodi/Component/Product/Repository/CategoryRepository.php
@@ -73,14 +73,14 @@ class CategoryRepository extends EntityRepository
     /**
      * Get the children categories given a parent category.
      *
-     * @param CategoryInterface $parentCategory The parent category.
+     * @param CategoryInterface|null $parentCategory The parent category.
      * @param bool              $recursively    If it should check the
      *                                          children recursively
      *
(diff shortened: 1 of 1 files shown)


rogergros:
Yep. I think it don't makes sense to ask for "Children categories" of a "null category". What it shoudl return? Empty array? Null? All the root categories?

There's a "getParentCategories" to get all the root categories and it don't makes sense to me to ask for children categories if you don't have a category.

claretcrab:
:+1: I will put this patch only in my project

Thanks!

mmoreram:
@claretcrab Could you please tell us why do you think this is wrong? Thanks.

claretcrab:
I'm not sure at a 100% but when I insert products with my bulk command, sometimes, when a category is inserted, bamboo/src/Elcodi/Admin/ProductBundle/EventListener/NewCategoryPositionEventListener.php fails because argument 1 is passed as null

I will keep it as backward compatibility, but I you are sure it won't fail with Bamboo, I'm fine to close this PR.

mmoreram:
@claretcrab Can you please ensure at the 100% of that? If there is an error I would like to know where it is instead of applying a non-valid patch. Thanks.

claretcrab:
100% it's a bug of my import

mmoreram:
@claretcrab ok

PR #785, opened 2015-05-13
Refactored Plugins and Menus

mmoreram:
- New implementation of Menus is based on
  - Filters - Return true or false if the node is discarted
  - Builders - Add new content into the menu
  - Modifiers - Modify actual content
- Any kind of these new classes can modify final menu and must be defined using tags
- When the menu is loaded from database is automatically detached from its entity manager
  to disallow the menu modification
- The implementation of Plugins is based on
  - Plugin yaml definition, including fields
  - A new formType is created given a plugin, in order to dynamically create the
    configuration page


src/Elcodi/Component/Menu/Modifier/MenuActiveModifier.php
```diff
@@ -0,0 +1,68 @@
...
+    {
+        $masterRequest = $this
+            ->requestStack
+            ->getMasterRequest();
+
+        if (!$masterRequest) {
+            return;
+        }
+
+        $currentRoute = $masterRequest->get('_route');
```
xphere: So, will this `modify` method run once for each node?
Even if it does not apply because there's no `$masterRequest`?

src/Elcodi/Component/Plugin/Repository/PluginRepository.php
```diff
@@ -15,13 +15,16 @@
...
+use Doctrine\ORM\EntityRepository;
 
 /**
- * Class PostMenuLoadEvent
+ * Class PluginRepository
  */
-class PostMenuLoadEvent extends AbstractMenuEvent
+class PluginRepository extends EntityRepository
 {
+    /**
```
xphere: This comments aren't needed.

src/Elcodi/Component/Plugin/Services/PluginManager.php
```diff
@@ -57,252 +69,194 @@ class PluginManager
...
+     * @return PluginInterface[]|Bundle[] Plugins installed
      */
-    public function getPlugins()
+    protected function getInstalledPluginBundles()
     {
-        if ($this->plugins) {
-            return $this->plugins;
-        }
+        $plugins = [];
+        $bundles = $this
```
xphere: If there's no other need for the `Kernel` to be injected, maybe better pass directly `kernel.bundles` as a parameter.

xphere:
It's hard to follow so many changes in a single commit.
I'd recommend to group related changes and split in multiple commits to ease the review.
Maybe a little bit of `git add --patch` will help.

src/Elcodi/Bundle/MenuBundle/Resources/config/services.yml
```diff
@@ -6,20 +6,21 @@ services:
     elcodi.manager.menu:
         class: Elcodi\Component\Menu\Services\MenuManager
         arguments:
-            - @elcodi.event_dispatcher.menu
-            - @elcodi.processor.menu_event
-            - @elcodi.serializer.menu
             - @elcodi.repository.menu
```
xphere: To me this looks like too many dependencies.

xphere:
I don't see the point that makes tagged services better than events for menu generation.

src/Elcodi/Component/Plugin/Repository/PluginRepository.php
mmoreram: Why?

src/Elcodi/Component/Menu/Modifier/MenuActiveModifier.php
mmoreram: Yes

src/Elcodi/Component/Plugin/Services/PluginManager.php
mmoreram: ok

src/Elcodi/Bundle/MenuBundle/Resources/config/services.yml
mmoreram: Ok, I'll split.

mmoreram:
I do. And because I've never been able to follow the logic using Events, and because an event should never be modifiable and return a value, using tags I think is the best practice.

Now we work with Entities, and is more clear and easy to understand/follow.

src/Elcodi/Component/Plugin/Repository/PluginRepository.php
mmoreram: Ups, yes

xphere: They don't apply anymore, maybe they were related to a former function that no longer exists.

PR #805, opened 2015-05-24
Fixed order factory when empty coupon

octante:
This PR is a fix of order factory when a new empty order is created. Error appears when load an order in admin and this order hasn't a coupon.

Order entity has a field couponAmount, and method getCouponAmount returns a Money object. To create this object we need an amount and a currency, and currency must be a CurrencyInterface object. If database field of currency is null Currency can be initialized and Money object is wrong.

I saw this error when I did a paypal purchase. Coupon field of database is filled in an event that is dispatched when payment is done, but if admin want to view this order in administration and paypal payment is not finished an error is showed.

To fix it I have created an empty coupon when an empty order is created. Another proposal is welcome :)

Thanks.

### DIFF
--- a/src/Elcodi/Bundle/CartBundle/Resources/config/factories.yml
+++ b/src/Elcodi/Bundle/CartBundle/Resources/config/factories.yml
@@ -21,6 +21,7 @@ services:
         arguments:
             - @elcodi.order_payment_states_machine_manager
             - @elcodi.order_shipping_states_machine_manager
+            - @elcodi.wrapper.default_currency
         calls:
             - [setEntityNamespace, ["%elcodi.entity.order.class%"]]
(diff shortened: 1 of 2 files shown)


mmoreram:
@octante Great.

The only point I disagree about this PR is that you're creating the new Money using `@elcodi.wrapper.currency`, but you should use the wrapper that provides you the default currency, called `@elcodi.wrapper.default_currency`.

octante:
Hi!!

ok, true! I made another PR with this change!

Thanks @mmoreram !! :D

mmoreram:
@octante Nopes! Please, reuse this PR with a `push --force` in order to have the historical conversation :) Is even easier for U indeed :)

Thanks!

octante:
I have used this PR :)

octante:
Updated :)

PR #812, opened 2015-06-03
Replace elcodi:~0.5.0 to elcodi:self.version

xphere:
Could something like this make sense? /ping @mmoreram 

This fixes a bug with `--prefer-lowest` when requiring a bundle `0.5.18` installs the component `0.5.0`.

### DIFF
--- a/src/Elcodi/Bundle/AttributeBundle/composer.json
+++ b/src/Elcodi/Bundle/AttributeBundle/composer.json
@@ -40,12 +40,12 @@
         "symfony/dependency-injection": "~2.6",
         "mmoreram/simple-doctrine-mapping": "~1.0",
 
-        "elcodi/core-bundle": "~0.5.0",
-        "elcodi/attribute": "~0.5.0"
+        "elcodi/core-bundle": "self.version",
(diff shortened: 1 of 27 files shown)


mmoreram:
Using this strategy, when your code is not on any version, then I think that `dev-master` is used instead.

xphere:
This is not using the root package version, but the root dependency to locate the underneath versions.
So, if your application depends on `elcodi/product-bundle: 0.5.17` and `elcodi/tax-bundle: 0.5.12`, you're depending also on `elcodi/product: 0.5.17` and `elcodi/tax: 0.5.12`.

xphere:
The problem is I don't know if we can enforce this retroactively to our old tags.

mmoreram:
Each version of Elcodi should only depend only on these versions that can be compatible with. Old versions of all packages are this. Old.

As soon as Elcodi reaches a `stable` version, then you can use the `^` element. Since this moment, old tags are wrong.

xphere:
Ok, so this makes sense only after reaching the 1.0 milestone. Makes sense.

PR #813, opened 2015-06-03
[WIP] Use master locale from settings

xphere:
When sorting active languages, it was promoting to master the wrong one (from `elcodi.locale`).
This uses the correct master-language from the configuration.

Possible improvements:
- Decouple from `elcodi/config`

### DIFF
--- a/src/Elcodi/Bundle/LanguageBundle/Resources/config/twig.yml
+++ b/src/Elcodi/Bundle/LanguageBundle/Resources/config/twig.yml
@@ -7,6 +7,6 @@ services:
         class: Elcodi\Component\Language\Twig\LanguageExtension
         arguments:
             - @elcodi.manager.language
-            - @elcodi.locale
+            - @=elcodi_config('store.default_language')
         tags:
             - { name: twig.extension }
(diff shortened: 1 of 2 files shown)


xphere:
This is wrong in many levels… needs to be moved to Bamboo, because it has no place in Elcodi.

xphere:
Ok, moving the master locale to a parameter of the function, this is now decoupled from `Configuration` and from the forthcoming `Store` entity.

But now I feel this is less useful, since we need to send an extra parameter in every twig template. Maybe there's need for a `LanguageWrapper`?

xphere:
Wow, now I see this is not a twig function, but a global! The method `getLanguages` is called just once, at twig initialization, so no way to customize from twig itself. I think we should find a better solution.

xphere:
I think it's better to leave this service as it is, and change the service definition in inherited projects (I'm thinking in Bamboo). Makes more sense? /ping @mmoreram

xphere:
Maybe something like this in Bamboo: https://gist.github.com/xphere/2780fd71d757df5110c9

xphere:
Deprecated by https://github.com/elcodi/bamboo/pull/491

PR #817, opened 2015-06-05
Store, Plugins, Shipping, Menu and more...

mmoreram:
## Store

* New store infrastructure, providing all store-related information in one single
  table and entity
* Removed configuration entries for old store information
* Fixed service definitions with new store services
* Fixed as well bundle dependencies (before ConfigurationBundle, now StoreBundle)
* There is a wrapper called `elcodi.wrapper.store` that loads current store.

## Plugins

* Removed Expression Language provider. Not necessary anymore as the bundle provides several
  ways of proper injection
* New plugin features
* Plugin management by console
    * elcodi.plugins.list
    * elcodi.plugin.enable
    * elcodi.plugin.disable
* If a plugin has a Resources/config/routing.yml file, then this file is loaded in kernel
* If a plugin has a Resources/config/external.yml file, then this file is prepend
* Different plugin types
    * plugin
    * template
* Plugin categories treated in a special way
    * payment
    * shipping

## Payment

* Created an engine for payment needs.
* Payment component and bundle will nicely ask everyone what payment methods are
  available and some extra information about how these payment methods should
  be displayed
* This component has it's own eventDispatcher and Event object

## Shipping

* Transformed Shipping engine from static platform to a simple engine
* Shipping engine is like Payment Engine. All logic related to custom
  shipping carriers will be placed in Bundles, so the component and the
  bundle will focus on the engine that will allow these Plugins to interact
  with the core.
* An event is dispatched when all shipping actors need to propose their prices,
  given a Cart
* All pugins can hook to this event in order to provide their options
* Removed Carrier and ShippingRange implementation. Placed as a Plugin in Bamboo

## EmptyMoneyWrapper

* In order to focus in one simple place the creation of a zero-valued money
  using the default currency, a new wrapper has been created

## Cart-related factories

* All Cart related factories
    * CartFactory
    * CartLineFactory
    * OrderFactory
    * OrderLineFactory
  will intialize their total prices with zero-valued money. Because these values
  are initialized on configuration time, can be the possibility that are not
  initialized on transformation.
* Using new EmptyMoneyWrapper

## Menu

* Added two nice features to menus
* By default, all modifiers (filters, changers and builders) will be applied to
  all menus, but once you define one of these using a DIC tag, you can define what
  menu/menus you want it to be applied to
* By default, all modifiers are applied before caching the menu, but you can define
  as well the tag with a stage value, being available `before_cache` and `after_cache`.
* isActive and isExpanded are now resolved dinamically, taking in account their children
  and the actual route
* Each menu node can have a priority, and when subnodes are selected, then all subnodes are
  returned already sorted by their priority
* Added new filter called MenuExpandedFilter. When a node has no children and has not url,
  then is removed.
* Modifiers for active or expanded have been removed

## Bundle Dependencies

* Bundle dependencies are now respecting the priority of the dependencies
* If A requires B, then B should be loaded before A
* Sometimes this rule cannot be solved (when cyclic dependencies)
* A dependency can be defined using the namespace of the bundle or with the instance itself

## EntityTranslator

* Fixed small error, visible when prepending new translation entities

## Location databases

* Added spain.sql and france.sql inside the GeoBundle

## Configuration

* Removed all configuration entries, in favor of Store implementation.


mmoreram:
@elcodi/owners Tachaaan!

One commit to rule them all. Sorry for this gorgeous commit, but I had no choice :(

PR #831, opened 2015-06-11
Address view

rogergros:
- Added Address view factory
- Added Address view

### DIFF
--- a/src/Elcodi/Component/Geo/View/Address.php
+++ b/src/Elcodi/Component/Geo/View/Address.php
@@ -1,0 +1,312 @@
+<?php
+
+/*
+ * This file is part of the Elcodi package.
+ *
+ * Copyright (c) 2014-2015 Elcodi.com
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * Feel free to edit as you please, and have fun.
+ *
+ * @author Marc Morera <yuhu@mmoreram.com>
+ * @author Aldo Chiecchia <zimage@tiscali.it>
+ * @author Elcodi Team <tech@elcodi.com>
+ */
+
+namespace Elcodi\Component\Geo\View;
+
+use DateTime;
+
+use Elcodi\Component\Geo\Entity\Interfaces\AddressInterface;
+use Elcodi\Component\Geo\Services\Interfaces\LocationProviderInterface;
+use Elcodi\Component\Geo\ValueObject\LocationData;
+use Elcodi\Component\Geo\View\Interfaces\AddressInterface as AddressViewInterface;
+
+/**
+ * Class Address
+ */
+class Address implements AddressViewInterface
+{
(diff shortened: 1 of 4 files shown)


src/Elcodi/Component/Geo/View/Address.php
```diff
@@ -0,0 +1,280 @@
...
+ * Class Address
+ */
+class Address implements AddressViewInterface
+{
+    /**
+     * @var array
+     *
+     * The hierarchies already retrieved.
+     */
+    private static $hierarchies = [];
```
mmoreram: If this class can  be overwritten, then this should be `protected`

src/Elcodi/Component/Geo/View/Address.php
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    /**
+     * Get Name
+     *
+     * @return mixed Name
+     */
+    public function getName()
+    {
+        return $this->address->getName();
```
mmoreram: return $this->address->getName();

src/Elcodi/Component/Geo/View/Address.php
```diff
@@ -0,0 +1,280 @@
...
+     */
+    public function getMobile()
+    {
+        return $this->address->getMobile();
+    }
+
+    /**
+     * Get Name
+     *
+     * @return mixed Name
```
mmoreram: Please, add good return types here

src/Elcodi/Component/Geo/View/Address.php
```diff
@@ -0,0 +1,280 @@
...
+     * Gets the country info
+     *
+     * @return LocationData
+     */
+    public function getCountryInfo()
+    {
+        $hierarchy = $this->getAddressHierarchy($this->address);
+
+        foreach ($hierarchy as $location) {
+            /** @var LocationData $location */
```
mmoreram: /**
 * @var LocationData $location
 */

src/Elcodi/Component/Geo/View/Address.php
rogergros: That's static because no matter how's created, this property should be shared. So it only checks the hierarchy once.

src/Elcodi/Component/Geo/View/Address.php
rogergros: Sorry, All these methods are copy pasted from the Address entity, I'll change it here.

src/Elcodi/Component/Geo/View/Address.php
rogergros: That's good. Mobile is optional so it can be null.

src/Elcodi/Component/Geo/View/Address.php
rogergros: I'll change it but is there any reason to avoid doing it in "one line"?

PR #852, opened 2015-06-15
Plugins are disabled by default but can be enabled

rogergros:
- Plugins are disabled by default
- You can configure a plugin to be enabled by default


src/Elcodi/Component/Plugin/Services/PluginConfigurationTree.php
```diff
@@ -60,6 +60,9 @@ public function getConfigTreeBuilder()
                 ->scalarNode('description')
                     ->defaultNull()
                 ->end()
+                ->scalarNode('enabled_by_default')
```
xphere: Not `scalarNode` but `booleanNode`.

PR #858, opened 2015-06-22
Make `PluginType` stateless

xphere:
As form types are resolved and cached, making them stateful could lead to some undesirable problems.
This makes the `PluginType` stateless by adding a new `plugin` option.

### DIFF
--- a/src/Elcodi/Bundle/PluginBundle/Resources/config/formTypes.yml
+++ b/src/Elcodi/Bundle/PluginBundle/Resources/config/formTypes.yml
@@ -5,3 +5,5 @@ services:
     #
     elcodi.form_type.plugin:
         class: Elcodi\Component\Plugin\Form\Type\PluginType
+        tags:
+            - { name: form.type, alias: elcodi_form_type_plugin }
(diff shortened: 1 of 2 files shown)


mmoreram:
@xphere what kind of problems?

xphere:
This is something @webmozart taught us in his Form workshop.
You can't make this form type a named one, because when the type is resolved and cached, the state is preserved, and you can't reuse it for another plugin in the same request.

Also, without named form types, you can't use the form type extension, for example. (That was my first concern)

mmoreram:
@xphere But you are adding this name to this FormType, right?

xphere:
Yes, but not only the name, I'm also solving the statefulness of the type by removing the field and start using the `OptionResolver`, which can be customized in each call.

xphere:
I can't update nor rebase this PR anymore. Should I close it?

mmoreram:
This PR has been merged into Shopery, right?
Then, the best procedure would be a PR from that repository, in order to give you the full control of the branch.

xphere:
Repulled in https://github.com/elcodi/elcodi/pull/868

PR #861, opened 2015-06-24
Added GD adapter for image resize

mmoreram:
* Set as default
* Changed the way these adapters are configured
* Fixed travis definition when composer update

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: php
 
 env:
-    - COMPOSER_OPTIONS="--prefer-source"
+    - COMPOSER_OPTIONS="install --prefer-source"
 
 php:
     - 5.4
@@ -14,7 +14,7 @@ matrix:
     fast_finish: true
     include:
         - php: 5.4
-          env: COMPOSER_OPTIONS="--prefer-lowest"
+          env: COMPOSER_OPTIONS="update --prefer-lowest"
 
 before_install:
     - sh -c "sudo mkdir vendor"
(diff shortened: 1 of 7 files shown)


alch:
:+1: we really needed this. The only thing i'd check is that it looks like it is possible to have divisions by zero. Maybe those cases are already covered by tests?

Class `Elcodi\Component\Media\Adapter\Resizer\Model\Dimensions`

[line 136](https://github.com/elcodi/elcodi/blob/master/src/Elcodi/Component/Media/Adapter/Resizer/Model/Dimensions.php#L136), [line 182](https://github.com/elcodi/elcodi/blob/master/src/Elcodi/Component/Media/Adapter/Resizer/Model/Dimensions.php#L182), [line 198](https://github.com/elcodi/elcodi/blob/master/src/Elcodi/Component/Media/Adapter/Resizer/Model/Dimensions.php#L198), [line 200](https://github.com/elcodi/elcodi/blob/master/src/Elcodi/Component/Media/Adapter/Resizer/Model/Dimensions.php#L200), [line 200](
https://github.com/elcodi/elcodi/blob/master/src/Elcodi/Component/Media/Adapter/Resizer/Model/Dimensions.php#L203), [line 204](https://github.com/elcodi/elcodi/blob/master/src/Elcodi/Component/Media/Adapter/Resizer/Model/Dimensions.php#L204)

mmoreram:
@alch All divisions by 0 are based on the idea that some data that should not be 0, is 0. So... the question... how you recommend it to be managed? Exceptions?

alch:
`\InvalidArgumentException` can be a solution. However I'd just open an issue and resolve it separately

PR #873, opened 2015-07-06
Removed property and access methods

mmoreram:
Non-used property is removed

### DIFF
--- a/src/Elcodi/Component/Cart/Entity/Interfaces/OrderInterface.php
+++ b/src/Elcodi/Component/Cart/Entity/Interfaces/OrderInterface.php
@@ -203,22 +203,6 @@ interface OrderInterface
     public function setWeight($weight);
 
     /**
-     * Get InvoiceAddress
-     *
-     * @return AddressInterface InvoiceAddress
-     */
-    public function getInvoiceAddress();
-
-    /**
-     * Sets InvoiceAddress
-     *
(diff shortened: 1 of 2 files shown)


mmoreram:
@rogergros @alch @xphere Can you please confirm that this property is not really used anywhere?

rogergros:
When was this created? :neutral_face: we have a "billingAddress" on cart that should be converted on the Cart to Order transformation process. 

Right now we are not generating invoices but in a future this will be needed to generate it, isn't it?

mmoreram:
@rogergros Yes, but for that we have TaxAddress, right?

rogergros:
@mmoreram Yes, that seems to be the case. Then it's ok for me :+1:

cordoval:
you still need to remove the methods in the interface:

```bash
...........PHP Fatal error:  Class Elcodi\Component\Cart\Entity\Order contains 2 abstract methods and must therefore be declared abstract or implement the remaining methods (Elcodi\Component\Cart\Entity\Interfaces\OrderInterface::getInvoiceAddress, Elcodi\Component\Cart\Entity\Interfaces\OrderInterface::setInvoiceAddress) in /home/travis/build/elcodi/elcodi/src/Elcodi/Component/Cart/Entity/Order.php on line 461
```

mmoreram:
Working on that ATM

cordoval:
:+1:

PR #885, opened 2015-08-10
Exchange rate calculator

rogergros:
- Added exchange rate calculator service
- This is usefull if you only need the exchange rate (e.g. to do masive changes)

### DIFF
--- a/src/Elcodi/Component/Currency/Services/CurrencyConverter.php
+++ b/src/Elcodi/Component/Currency/Services/CurrencyConverter.php
@@ -31,34 +31,22 @@ use Elcodi\Component\Currency\Exception\CurrencyNotConvertibleException;
 class CurrencyConverter
 {
     /**
-     * @var boolean
+     * @var ExchangeRateCalculator
      *
-     * Multiply
+     * The exchange rate calculator.
      */
-    const MULTIPLY = true;
-
-    /**
-     * @var boolean
-     *
-     * Multiply
-     */
-    const DIVIDE = false;
-
-    /**
-     * @var CurrencyManager
-     *
-     * Currency manager
-     */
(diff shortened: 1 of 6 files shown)


src/Elcodi/Component/Currency/Services/CurrencyConverter.php
```diff
@@ -113,81 +101,14 @@ private function convertCurrency(
             return Money::create($amount, $currencyFrom);
         }
 
-        $convertedAmount = $this->convertBetweenIsos(
-            $currencyFrom->getIso(),
-            $currencyTo->getIso(),
-            $amount
+        $exchangeRate = $this->exchangeRateCalculator->calculateExchangeRate(
```
mmoreram: Can you format this line with new lines?

rogergros: Done!

mmoreram:
Could you please explain us what this change is intended for? There are some important changes here!

rogergros:
As I pointed you sometimes could need the exchange rate to do certain actions like bulk products currency change. Right now it's not possible to get the Exchange rate for a currency.

Besides this there's also a coupling with the USD (It's harcoded in some points of the code) while most of the code uses the currency parameter and not the USD (Like the currency populator). This could lead to a bit problem. E.g. I'm using the EUR as default money, I populate the DB and the populator gets all the exchange rates from EUR to all my currencies but the MoneyConvertor uses the USD (Harcoded) to exchange the money.

mickaelandrieu:
:+1:

mmoreram:
need some time for reviewing. Thanks @rogergros

rogergros:
:+1:

PR #896, opened 2015-09-08
Can disable entities

xphere:
Elcodi entities can't be disabled, because `cannotBeEmpty` don't let you to set `enabled` to `false`.
Removing the call fixes this bug.

Also, adding `treatFalseLike` in the root node, you can disable an entity easier.
Before:
```yml
elcodi_banner:
    mapping:
        banner:
            enabled: false
        banner_zone:
            enabled: false
```

After:
```yml
elcodi_banner:
    mapping:
        banner: false
        banner_zone: false
```


mmoreram:
YAY! Great @xphere +kudos!

PR #909, opened 2015-09-28
Casting and missing types requirement

EmanueleMinotto:
Things I'm doing in this PR:

* adding missed parameter types (ex: `AddressInterface` into src/Elcodi/Component/Cart/Entity/Cart.php)
* `use` statements for repeated FQCN
* parameters casting

Why parameters casting? Imo could be useful for different reasons:

1. if the parameter implements the `__toString` method, than it can be used without breaking internal code
2. a basic validation system ([in just one opcode](https://3v4l.org/UIWWv/vld#tabs), maximum 3 for assignments)

### DIFF
--- a/src/Elcodi/Component/Banner/Services/BannerManager.php
+++ b/src/Elcodi/Component/Banner/Services/BannerManager.php
@@ -57,7 +57,7 @@ class BannerManager
     {
         return $this
             ->bannerRepository
-            ->getBannerByZone($bannerZoneCode, $language);
+            ->getBannerByZone((string) $bannerZoneCode, $language);
     }
 
     /**
--- a/src/Elcodi/Component/Banner/Entity/Banner.php
+++ b/src/Elcodi/Component/Banner/Entity/Banner.php
@@ -88,7 +88,7 @@ class Banner implements BannerInterface
      */
     public function setName($name)
     {
-        $this->name = $name;
+        $this->name = (string) $name;
 
         return $this;
     }
@@ -112,7 +112,7 @@ class Banner implements BannerInterface
      */
     public function setDescription($description)
     {
-        $this->description = $description;
+        $this->description = (string) $description;
 
         return $this;
     }
@@ -136,7 +136,7 @@ class Banner implements BannerInterface
      */
     public function setUrl($url)
     {
-        $this->url = $url;
+        $this->url = (string) $url;
 
         return $this;
     }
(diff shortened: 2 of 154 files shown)


src/Elcodi/Component/Banner/Services/BannerManager.php
```diff
@@ -57,7 +57,7 @@ public function getBannersFromBannerZoneCode($bannerZoneCode, LanguageInterface
     {
         return $this
             ->bannerRepository
-            ->getBannerByZone($bannerZoneCode, $language);
+            ->getBannerByZone((string) $bannerZoneCode, $language);
```
mmoreram: Hmmm, be careful here. This casting is pretty interesting if it is done only at the model.
What do you think?

EmanueleMinotto: Imo where there's an abstraction there should be a "trust level".
If the `bannerRepository` is always trusted, than no problem, everything can be passed to it, but if the `bannerRepository` isn't so "strong" to accept every type and requires a string (based on `BannerManager` docblock) than a string must be passed.

I'm not sure if I explained well what I'm thinking ^^'

EmanueleMinotto:
`Elcodi\Component\Language\Entity\Locale` doesn't accept only strings, should the docblock be updated?

```
Elcodi\Component\Currency\Tests\UnitTest\Services\MoneyPrinterTest::testCurrencyRateNotFoundThrowsException
Object of class Mock_LocaleInterface_c74ca7ab could not be converted to string

elcodi/src/Elcodi/Component/Language/Entity/Locale.php:41
elcodi/src/Elcodi/Component/Language/Entity/Locale.php:63
elcodi/src/Elcodi/Component/Currency/Tests/UnitTest/Services/MoneyPrinterTest.php:102
```

Tried to cast `$localeIso` to string (line 41)

mmoreram:
Tests are not passing. I am reviewing this PR right now :)
I'll ping you then!

Thanks!

mmoreram:
That's a lot of work, man!
There's a lot of philosophy and definition here I would like to talk about... this PR is too big, but I like a lot the idea behind it.

Some points to discuss about here:

## Issues and minor fixes

There are some issues fixed here, for example when you cast some complex objects in entities (Collection, AddressInterface...). These fixes deserve a unique commit, so it can be merged without discussion.

## Improvements in trust layer

You're right, services trust is something we should talk about.
When you create a new object with a PHPDoc, then you're creating each time a contract between the user and values are being passed to the service itself. This is done each time a class uses another class in our service layer.

So, said that, casting the objects is something you should only do once you manipulate a value from a variable (controllers, commands...). Doing it always means that you don't really trust in this contract, so then, something is wrong here.

We should add a chapter in the documentation about this trusting layer... seems really interesting to talk about that.

PR #916, opened 2015-10-10
added .gitattributes

EmanueleMinotto:
|Q            |A  |
|---          |---|
|Fixed Tickets|   |
                   

Considering CI systems, I think some files (including the *Tests* directory) should be removed when developers download dependencies using `--prefer-dist`.

thephpleague is doing this: https://github.com/thephpleague/skeleton/blob/master/.gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,16 @@
+# Path-based git attributes
+# https://www.kernel.org/pub/software/scm/git/docs/gitattributes.html
+
+# Ignore all test and documentation with "export-ignore".
+.editorconfig       export-ignore
+.formatter.yml      export-ignore
+.gitattributes      export-ignore
+.gitignore          export-ignore
+.gitsplit.yml       export-ignore
+.gush.yml           export-ignore
+.php_cs             export-ignore
+.travis.yml         export-ignore
+composer.lock       export-ignore
+DataFixtures        export-ignore
+phpunit.xml.dist    export-ignore
+Tests               export-ignore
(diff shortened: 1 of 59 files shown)


javiereguiluz:
Beware that other important PHP projects, like Symfony, have rejected this idea because of the problems that it introduces. Reference: https://github.com/symfony/symfony/issues?utf8=%E2%9C%93&q=is%3Aissue+gitattributes

EmanueleMinotto:
I agree about not including the .gitattributes in symfony/symfony, but this is a different case imo.
Starting from https://github.com/symfony/symfony/issues/6605:

> people use some of our test classes in their own tests (we have already moved some of them to a Test/ directory and that helps bu I fear that many more will need to be moved -- see https://twitter.com/huizendveld/status/288331330998116352 for an example);

elcodi/elcodi is at a higher level than symfony, I don't think that elcodi should be used for tests except for the [TestCommonBundle](https://github.com/elcodi/TestCommonBundle) (not included in this PR).

mmoreram:
Not sure about that yet...

mmoreram:
I've made some research, and removing all these fields have been a pain for me sometimes. Having files, even if they don't do anything useless, i not bad. Maybe sometimes you need to do some development on top of that status, and then you are f***d.

Tests are great always, so help people with the comprehension of the project itself.
I would not consider removing them never (they are part of the project itself)

Thanks for the effort of adding this configuration, but I have to decline for now :/

Issue #918, opened 2015-10-11
Add composer commands


EmanueleMinotto:
A fixed version of the php-cs-fixer producing a different output from the one produced by a more recent version could be confusing (don't know if this is only for me or not ^^'), so my proposal is to use composer commands, for example something like `composer check-style`.

Advantages:

1. the project version will be used, so if the user has different global and project versions, with that command the project version is used
2. more commands can be grouped, see the example below

```json
{
    "scripts": {
        "check-style": [
            "php-cs-fixer fix src",
            "php-formatter f:u:s src"
        ]
    }
}
```

Of course this example could be applied also to tests if fixed versions are required.

PR #919, opened 2015-10-12
added composer scripts

EmanueleMinotto:
Closes https://github.com/elcodi/elcodi/issues/918

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -148,5 +148,13 @@
         "branch-alias": {
             "dev-master": "1.0-dev"
         }
+    },
+    "scripts": {
+        "check-style": [
+            "php-cs-fixer fix --config-file=.php_cs",
+            "php-formatter f:h:f src",
+            "php-formatter f:u:s src"
+        ],
+        "test": "phpunit"
     }
 }
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,12 +6,14 @@ that your code fulfills these standards before any Pull Request.
 
 ``` bash
 composer.phar update
-bin/php-cs-fixer fix
-bin/php-formatter formatter:use:sort src/
-bin/php-formatter formatter:header:fix src/
+composer.phar check-style
 ```
 
 There is also a policy for contributing to this project. Pull requests must
 be explained step by step to make the review process easy in order to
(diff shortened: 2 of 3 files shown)


EmanueleMinotto:
Some dependies now require php >= 5.5
@mmoreram should I freeze those dependencies to be compatible with php 5.4?

mmoreram:
@EmanueleMinotto What do U mean about requirement of php >= 5.5 ?

mmoreram:
Oh... this new feature requires PHP 5.5?
We haven't changed the `composer.json` definition!
Why is happening this?

mmoreram:
@EmanueleMinotto BTW, we cannot update the minimum PHP version... I mean, we could, but it has no sense when our mission is making this instalable as easy as possible...

EmanueleMinotto:
No, this feature is based on Composer, I mean that some dependencies like [zendframework/zend-eventmanager](https://packagist.org/packages/zendframework/zend-eventmanager#2.5.2) (see the composer.lock) now require php 5.5.
I can freeze them to the last version availabe for old php versions (5.3 and 5.4), but some of them won't upgrade doing this (for example zendframework/zend-eventmanager did this switching from 2.5.1 to the 2.5.2).

EmanueleMinotto:
Solved the dependencies problem updating vendor by vendor.

composer.json
```diff
@@ -148,5 +148,13 @@
...
             "dev-master": "1.0-dev"
         }
+    },
+    "scripts": {
+        "check-style": [
+            "php-cs-fixer fix --config-file=.php_cs",
+            "php-formatter f:h:f src",
+            "php-formatter f:u:s src"
+        ],
+        "test": "phpunit"
```
mmoreram: not phpunit, but bin/phpunit, right?

EmanueleMinotto: No because `bin/` files are executed first.

https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands :
> Note: Composer's bin-dir is pushed on top of the PATH so that binaries of dependencies are easily accessible as CLI commands when writing scripts.

mmoreram: Great :dancer:

PR #930, opened 2015-10-15
added php versions matrix

EmanueleMinotto:
|Q            |A  |
|---          |---|
|Fixed Tickets|   |

Added current PHP releases available on http://windows.php.net/download/


mmoreram:
Great :D

PR #932, opened 2015-10-16
types casting

EmanueleMinotto:
Related to https://github.com/elcodi/elcodi/pull/909

This PR includes strict parameters casting.

### DIFF
--- a/src/Elcodi/Component/User/Entity/Interfaces/LastLoginInterface.php
+++ b/src/Elcodi/Component/User/Entity/Interfaces/LastLoginInterface.php
@@ -17,6 +17,8 @@
 
 namespace Elcodi\Component\User\Entity\Interfaces;
 
+use DateTime;
+
 /**
  * Interface LastLoginInterface
  */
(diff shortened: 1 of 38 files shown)


src/Elcodi/Component/User/Entity/Interfaces/LastLoginInterface.php
```diff
@@ -36,5 +36,5 @@ public function getLastLoginAt();
      *
      * @return $this Self object
      */
-    public function setLastLoginAt($lastLoginAt);
+    public function setLastLoginAt(\DateTime $lastLoginAt);
```
mmoreram: We should avoid complete namespaces here. Instead of that, we should use here `DateTime` here, adding as well the use statemenet on the top of the file (Everywhere but Traits, remember)

mmoreram:
This PR is very important, because will carry some BC Breaks.
So, with the right communication we can do that.
I would add to this PR some extra things:
* Adding the null default value to these Object setters that can handle null values

EmanueleMinotto:
Moved to https://github.com/elcodi/elcodi/pull/943

Issue #942, opened 2015-10-26
Splitting documentations and prevent maintainance


EmanueleMinotto:
Documentations (coverage & docs) are currently included only in elcodi/elcodi and related repositories like elcodi/User don't contain any of them.

I would suggest another strategy to prevent thier maintainance and a not required inclusion inside the source branch while keeping them updated with the source code:
![strategy](http://www.websequencediagrams.com/cgi-bin/cdraw?lz=TG9jYWxob3N0LT5HaXRIdWI6IGdpdCBwdXNoCgALBi0-Z2l0c3BsaXQ6IAACBXRpbmcgc3ViZGlyZWN0b3JpZXMKABsIACsbVHJhdmlzLUNJOiBwb3N0LXB1c2ggaG9vawoAEQkAGA1hZnRlcl9zY3JpcHQgAAEGAB4MAIEmCABCBWRvY3MgaW50byBnaC1wYWdlcwo&s=mscgen)

The required lines in `.travis.yml` are:

```yaml
php:
- 5.6

env:
  global:
    secure: *travis secure key*

after_script:
- if [ $TRAVIS_PHP_VERSION = '5.6' ]; then sh generate-doc.sh; fi
```

and in the *generate-doc.sh* file the only required variable (extracted from env) is `${GH_TOKEN}`.

mmoreram:
Travis do that? I mean, does Travis can build things like that?

EmanueleMinotto:
Yes, check https://github.com/EmanueleMinotto/TwigCacheBundle
The related [API documentation](http://emanueleminotto.github.io/TwigCacheBundle/) is generated automatically using the [generate-api.sh](https://github.com/EmanueleMinotto/TwigCacheBundle/blob/master/generate-api.sh) and [.travis.yml](https://github.com/EmanueleMinotto/TwigCacheBundle/blob/master/.travis.yml) files.

PR #949, opened 2015-11-02
Coverage and documentation

EmanueleMinotto:
Closes https://github.com/elcodi/elcodi/issues/942

I've removed the folders *docs* and *coverage* in favour of this system for the advantages described in the related issue.

What's changed:

1. removed the `COMPOSER_OPTIONS` env variable since it's no more used and replaced the command with `composer install` (there's no composer.lock in the repository)
2. removed the mount commands, travis works without them too
3. added the code coverage using [coveralls.io](https://coveralls.io/), this allows the scrutinizer analysis to be fast as now
   1. to upload coverage results generated by phpunit, the library `satooshi/php-coveralls` is needed, so I'm using it only on travis
   2. `satooshi/php-coveralls` works only if executed from `vendor/bin/coveralls` and considering that `"bin-dir": "bin"` isn't required (nor an utility like could on elcodi/bamboo for Symfony v3) I've restored it to the default value
4. code coverage requires xdebug, that slows down the testing very much, so it's executed only if it's not a pull request
   1. actually is executed also on hhvm and php 7 even without producing a result, when there will be a support also for them the `.travis.yml` should be updated removing blocking conditions
5. the documentation is moved to the *gh-pages* branch and works only for the main repository (let me know if you want do the same also for subsplitted repositories)
   1. the documentation is generated using [Sami](https://github.com/FriendsOfPHP/Sami), but the file was moved from *.sami* to *sami.php* because *.sami* wasn't accepted as argument (I don't know the reason)
   2. the documentation is generated by travis and the file `generate-api.sh` only if it's not a pull request, if it's the the currently active PHP version (5.6 in this moment) and something should be changed once accepted this PR:
      * the `$TRAVIS_BRANCH = "master"` condition about the branch must be added to `.travis.yml` into the documentation generation line
      * the right repository must be changed at `git remote add origin https://${GH_TOKEN}@github.com/[repository].git > /dev/null`
   3. the documentation generation and access from travis require a crypted access token, see below

To generate the access token:
``` bash
$ curl -X POST -u <your_github_username> -H "Content-Type: application/json" -d "{\"scopes\":[\"public_repo\"],\"note\":\"token for pushing from travis\"}" https://api.github.com/authorizations
```

Once obtained the access token, to add the crypted secure key use:
``` bash
gem install travis
travis encrypt -r <user>/<repository> GH_TOKEN=<token> --add env.global
```

The result of this PR is available at:
 * **Code Coverage**: https://coveralls.io/github/EmanueleMinotto/elcodi
 * **API Documentation**: http://emanueleminotto.github.io/elcodi/


mmoreram:
Amazing man... amazing...

EmanueleMinotto:
The PR should be complete now.

PR #956, opened 2015-11-03
Issuse #921 All events set final classes

mmoreram:
* This change introduces BC Break somehow, only for people that have extended elcodi
  events.
* This binds the domain events, and protects it for extension


jontorrado:
What is the reason for protecting domain events?

theofidry:
immutability and final classes just :heart:

mmoreram:
@jontorrado We want to prevent anyone extending our domain events. No one should use them indeed, they should only be created in our event dispatchers.

mmoreram:
@EmanueleMinotto do I have your approval with this? :dancers:

EmanueleMinotto:
@mmoreram yes, after the explanation on Gitter this is ok for me too :+1:

PR #959, opened 2015-11-03
Moved setup and teardown before and after class

EmanueleMinotto:
Split of https://github.com/elcodi/elcodi/pull/958

Database schema generation and fixtures were executed for every test, this slows down the execution a lot, so I moved database related stuffs (setup and teardown) before and after class so some of them could be prevented.

For tests requiring explicitly everything new I've included:
```php
static::setUpBeforeClass();
$this->setUp();
```


mmoreram:
Moved to #961

PR #963, opened 2015-11-04
phpdoc fixes + attributes initialization

mmoreram:


src/Elcodi/Component/Plugin/EventDispatcher/EventAdapter.php
```diff
@@ -33,7 +33,7 @@ class EventAdapter extends Event implements EventInterface
      *
      * Context from the caller
      */
-    private $context;
+    private $context = [];
```
mmoreram: Not necessary... is private and the constructor set the final value

src/Elcodi/Component/Plugin/Entity/PluginConfiguration.php
```diff
@@ -29,7 +29,7 @@ class PluginConfiguration
      *
      * Encoded configuration
      */
-    protected $configuration;
+    protected $configuration = [];
```
mmoreram: Not necessary... the constructor sets the value

src/Elcodi/Component/Metric/API/Twig/APIMetricExtension.php
```diff
@@ -123,7 +123,7 @@ public function getAccumulation(
      * @param string   $event Token
      * @param string[] $dates Dates
      *
-     * @return array Distribution values
+     * @return integer Distribution values
```
mmoreram: Why? This method return an array

PR #972, opened 2015-11-05
Placed all logic in service layer

mmoreram:
* Twig extension had some logic inside. This logic has been extracted from there and placed in the service layer
* Added some tests for new services (functional in bundles, unit in components)
* Added these new services in the Symfony DIC layer (bundles)
* Added missing functional tests for all services (just testing the service requirement itself)

### DIFF
--- a/src/Elcodi/Component/Product/Services/PurchasableNameResolver.php
+++ b/src/Elcodi/Component/Product/Services/PurchasableNameResolver.php
@@ -1,0 +1,72 @@
+<?php
+
+/*
+ * This file is part of the Elcodi package.
+ *
+ * Copyright (c) 2014-2015 Elcodi Networks S.L.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * Feel free to edit as you please, and have fun.
+ *
+ * @author Marc Morera <yuhu@mmoreram.com>
+ * @author Aldo Chiecchia <zimage@tiscali.it>
+ * @author Elcodi Team <tech@elcodi.com>
+ */
+
+namespace Elcodi\Component\Product\Services;
+
+use Elcodi\Component\Attribute\Entity\Interfaces\ValueInterface;
+use Elcodi\Component\Product\Entity\Interfaces\ProductInterface;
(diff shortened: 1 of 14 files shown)


mmoreram:
@EmanueleMinotto the last one :D (rebased as well)

src/Elcodi/Component/Product/Services/PurchasableNameResolver.php
```diff
@@ -0,0 +1,68 @@
...
+                 * @var ValueInterface $option
+                 */
+                $productName .= $separator .
+                    $option->getAttribute()->getName() .
+                    ' ' .
+                    $option->getValue();
+            }
+        }
+
+        return $productName;
```
EmanueleMinotto: this can be replaced with:

```php
        if ($purchasable instanceof ProductInterface) {
            return $purchasable->getName();
        }

        /**
         * @var VariantInterface $purchasable
         */
        $productName = $purchasable->getProduct()->getName();

        foreach ($purchasable->getOptions() as $option) {
            /**
             * @var ValueInterface $option
             */
            $productName .= $separator .
                $option->getAttribute()->getName() .
                ' ' .
                $option->getValue();
        }

        return $productName;
```

src/Elcodi/Component/Product/Services/PurchasableNameResolver.php
```diff
@@ -0,0 +1,68 @@
...
+     * type of purchasable element.
+     *
+     * @param PurchasableInterface $purchasable Purchasable to get name from
+     * @param string               $separator   Separator string for product variant options
+     *
+     * @return string Purchasable name
+     */
+    public function getPurchasableName(
+        PurchasableInterface $purchasable,
+        $separator = ' - '
```
EmanueleMinotto: What about a `const DEFAULT_SEPARATOR = ' - ';` and here `$separator = self::DEFAULT_SEPARATOR`?
In this way the default separator can be identified outside of this class.

mmoreram:
@EmanueleMinotto updated with your ideas updated :)

EmanueleMinotto:
now it's good for me, well done :+1:

Issue #974, opened 2015-11-05
Fix APIMetricExtension methods visibility


mmoreram:
Try to remove all methods and work only with anonymous functions

PR #975, opened 2015-11-05
Issue #974 Created anonymous functions

mmoreram:
* Instead of public functions.

### DIFF
--- a/src/Elcodi/Component/Metric/API/Twig/APIMetricExtension.php
+++ b/src/Elcodi/Component/Metric/API/Twig/APIMetricExtension.php
@@ -52,87 +52,31 @@ class APIMetricExtension extends Twig_Extension
     public function getFunctions()
     {
         return [
-            new Twig_SimpleFunction('metric_beacon_unique', [$this, 'getBeaconsUnique']),
-            new Twig_SimpleFunction('metric_beacon_total', [$this, 'getBeaconsTotal']),
-            new Twig_SimpleFunction('metric_accumulation', [$this, 'getAccumulation']),
-            new Twig_SimpleFunction('metric_distributions', [$this, 'getDistributions']),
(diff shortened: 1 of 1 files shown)


src/Elcodi/Component/Metric/API/Twig/APIMetricExtension.php
```diff
@@ -52,87 +52,31 @@ public function __construct(AbstractMetricsBucket $metricBucket)
...
-     */
-    public function getDistributions(
-        $token,
-        $event,
-        $dates
-    ) {
-        return (int) $this
-            ->metricBucket
-            ->getDistributions($token, $event, $dates);
+                return (int) $this
```
EmanueleMinotto: `getDistributions` return an array

src/Elcodi/Component/Metric/API/Twig/APIMetricExtension.php
```diff
@@ -52,87 +52,31 @@ public function __construct(AbstractMetricsBucket $metricBucket)
...
-    public function getAccumulation(
-        $token,
-        $event,
-        $dates
-    ) {
-        return $this
-            ->metricBucket
-            ->getAccumulation($token, $event, $dates);
-    }
+                return $this
```
EmanueleMinotto: `getAccumulation` return an integer, so can be forced like in the closures above

EmanueleMinotto:
@mmoreram after those two fixes, this PR will be ok for me :+1:

mmoreram:
@EmanueleMinotto If removed all logic from the extension, and this includes all casting. When you add any `(int)` then you're creating logic, but the responsibility to add this changes is from the service itself. I've removed all logic :)

EmanueleMinotto:
seems right, well done :)

PR #978, opened 2015-11-10
Added interfaces for some final classes

EmanueleMinotto:
Acutally the final events are causing some problems because can't be mocked by [PHPUnit](https://phpunit.de/manual/current/en/test-doubles.html) nor by [PhpSpec](https://github.com/phpspec/prophecy/issues/197), generally final classes can't be mocked.

As suggested [here](http://ocramius.github.io/blog/when-to-declare-classes-final/) I've implemented an interface for the classes those are causing this problem (excluding the `__construct` method to not force implementations).


mmoreram:
As we spoke on gitter, having interfaces in events has not sense.
Events are not classes that can be implemented by others, so having interfaces here is not the right way of solving some problems.

Let's solve issues in Bamboo related to final classes.

@EmanueleMinotto agree?

EmanueleMinotto:
Yep :+1:

Issue #894, opened 2015-08-29
[Payment] Add payment method in Order


mmoreram:
This change affects an interface (Elcodi uses Header Interfaces), so that means that this change will be done in Elcodi 2.0

PR #982, opened 2015-11-18
Issue #894 Added payment method in Order

mmoreram:
* Added as well PaymentMethod extra data in order
* Added as well ShippingMethod extra data in order


src/Elcodi/Component/Cart/Entity/Order.php
```diff
@@ -106,6 +107,27 @@ class Order implements OrderInterface
     protected $shippingMethod;
 
     /**
+     * @var string
```
EmanueleMinotto: Why are these attributes strings?

src/Elcodi/Component/Cart/Entity/Order.php
```diff
@@ -106,6 +107,27 @@ class Order implements OrderInterface
...
+     *
+     * Shipping method extra
+     */
+    protected $shippingMethodExtra;
```
EmanueleMinotto: Initialize them as arrays allows you to remove these last two lines: https://github.com/elcodi/elcodi/pull/982/files#diff-9c57f24f614243f7a78090b60f906198R99

Issue #986, opened 2015-11-19
Problem with coupons usage


kolmanpav:
Usage of coupons (field in model) is not incremented. I found that in ElcodiCouponBundle / DependencyInjection/ ElcodiCouponExtension.php in getConfigFiles method is not eventListeners.yml file. That's why MakeCouponUsedEventListener is not correctly registered and incrementing of coupon usage doesn't work correctly

mmoreram:
We will use this issue to resolve several issues related to these bundles
* CouponBundle + Coupon
* CartCouponBundle + CartCoupon

This set of issues is because I've noticed that some parts of that application are difficult to follow. For this reason, we can just take a look at both bundles and components, and make some movements in order to clean a little bit the packages.

- [ ] Resolve the issue itself
- [ ] Move all business logic from EventListeners to Services
- [ ] Check the workflow when a Coupon is added into a Cart, test it all and document

This is all about consistency. Each time a Coupon is added into a Cart, all data should be updated properly. We could define all possible workflows here (for documentation as well), and test them all. Having all logic inside service will make much easier unit test, so listener should not be tested.

- [ ] Coupon has a field called `$absolutePrice`.

This value is not saved to database, but is loaded from the Cart price (for example, if the Coupon is 50% and the Cart is 10€, then this value is 5€). This is not well designed, so a Coupon should be part of two carts at the same.

We should place this field in CartCoupon. The converted value should be placed in OrderCoupon.
This task is about defining tests, and do the refactor.

I will add this task for `v1.2` because needs some work, and we should do it properly, in order to not create much BC Break (using deprecations instead of removing)

PR #997, opened 2015-11-23
Issue #986 Isolated stackable business logic

mmoreram:
* Removed logic from event listener and added to a service
* Tested class with unit test
* Tested event listener with functional tests

### DIFF
--- a/src/Elcodi/Component/CartCoupon/Tests/Services/StackableCouponCheckerTest.php
+++ b/src/Elcodi/Component/CartCoupon/Tests/Services/StackableCouponCheckerTest.php
@@ -1,0 +1,114 @@
+<?php
+
+/*
+ * This file is part of the Elcodi package.
+ *
+ * Copyright (c) 2014-2015 Elcodi Networks S.L.
+ *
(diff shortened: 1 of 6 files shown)


src/Elcodi/Component/CartCoupon/Tests/Services/StackableCouponCheckerTest.php
```diff
@@ -0,0 +1,114 @@
...
+     */
+    public function dataCheckStackableCoupon()
+    {
+        return [
+            [true, [], false],
+            [false, [], false],
+            [false, [true, true], true],
+            [false, [false, true], true],
+            [true, [false, true], true],
+            [true, [true, true], false],
```
EmanueleMinotto: Imo adding a comment here for every case could be useful for the future. :)

mmoreram: Updated to both tests data providers :)

EmanueleMinotto:
this is great, congrats!

mmoreram:
@EmanueleMinotto Thanks :dancer:

PR #998, opened 2015-11-26
Big refactor to CartCoupon layer

mmoreram:
## Component
- Removed all data from the event listener layer and moved to service layer
- Some service have been created, some of them have been moved (name changed)
- All services now have more intuitive names
- EventListeners have changed their names and definitions. Only one parameter in the constructor.
- The name of all EventListeners is "doSomething" and the method the same than the service method is called
- Some services have been covered with UnitTesting
- We have removed all unit testing from event listeners, because has not logic inside
- All event listeners are final now (they are never mocked, nor should be overwritten at all)
- Exceptions are final as well for the same reason

## Bundles
- Removed all service testing, just testing the application itself (a bundles is only a application layer)
- All event listeners covered with functional tests

## About the BC Break.
- Because this bundle/component is not required by any other component in all the application, then this can work
- Bamboo will need updating of version (in that case, yes, BCBreak somehow (we need to check)


src/Elcodi/Component/CartCoupon/Entity/CartCoupon.php
```diff
@@ -74,20 +84,10 @@ public function getCart()
      *
      * @return $this Self object
      */
-    public function setCoupon(CouponInterface $coupon)
+    public function setCoupon($coupon)
```
EmanueleMinotto: Why did you remove the `CouponInterface` here?

src/Elcodi/Component/CartCoupon/Entity/CartCoupon.php
```diff
@@ -44,13 +44,23 @@ class CartCoupon implements CartCouponInterface
...
+
+    /**
      * Sets Cart
      *
      * @param CartInterface $cart Cart
      *
      * @return $this Self object
      */
-    public function setCart(CartInterface $cart)
+    public function setCart($cart)
```
EmanueleMinotto: Why did you remove the `CartInterface` here?

mmoreram: LOL... :/

src/Elcodi/Component/CartCoupon/Entity/CartCoupon.php
mmoreram: same LOL here :/

PR #1004, opened 2015-11-27
Added new dependency with library

mmoreram:
* Changed usage of this feature to new one
* Old implementation should be keeped for old users, being permisive (this means
  allowing new interface and old at the same time)
* Deprecated old classes
* Removed tests of old implementation


EmanueleMinotto:
:+1:

PR #1006, opened 2015-11-27
Exclude tests from production classes map

EmanueleMinotto:
The Composer production classmap generator will ignore all files in
these paths. The paths are absolute from the package root directory,
`**` is implicitly added to the end of the paths.

Ref: https://getcomposer.org/doc/04-schema.md#exclude-files-from-classmaps

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -138,6 +138,9 @@
         "phpunit/phpunit": "4.5.0"
     },
     "autoload": {
+        "exclude-from-classmap": [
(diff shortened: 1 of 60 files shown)


EmanueleMinotto:
@mmoreram do you think this feature should be tested too?

mmoreram:
Well, you cannot test something in a production environment XD

mmoreram:
Well done :D

Issue #1008, opened 2015-11-28
Menu node issers


EmanueleMinotto:
The `Elcodi\Component\Menu\Entity\Menu\Node` methods:
* `isActive`
* `isExpanded`

have some problems if the Node isn't initialized with some data.

mmoreram:
When Node is initialized with the factory, then isExpanded should work as expected

mmoreram:
working on that

PR #1012, opened 2015-11-30
Added a layer of checking on Node

mmoreram:
* Added some tests covering these methods

### DIFF
--- a/src/Elcodi/Component/Menu/Entity/Menu/Node.php
+++ b/src/Elcodi/Component/Menu/Entity/Menu/Node.php
@@ -17,6 +17,8 @@
 
 namespace Elcodi\Component\Menu\Entity\Menu;
 
+use Doctrine\Common\Collections\Collection;
+
 use Elcodi\Component\Core\Entity\Traits\EnabledTrait;
 use Elcodi\Component\Core\Entity\Traits\IdentifiableTrait;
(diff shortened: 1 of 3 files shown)


mmoreram:
@EmanueleMinotto related to #1008

src/Elcodi/Component/Menu/Entity/Menu/Node.php
```diff
@@ -290,6 +297,10 @@ public function isActive($currentUrl)
      */
     public function isExpanded($currentUrl)
     {
+        if (!($this->subnodes instanceof Collection)) {
```
EmanueleMinotto: You could also remove the internal parentheses: `if (!$this->subnodes instanceof Collection) {`

mmoreram: Are you sure of that?
first is `!$this->subnodes`
then `{result} instanceof Collection

EmanueleMinotto: check the VLD: https://3v4l.org/MJuCR/vld#tabs
the `instanceof` has a higher precedence than `!` and both of them give the same output

mmoreram: Good then :D Thanks for that info! ^^ I was very used of working with these parenthesis :dancer:

PR #1013, opened 2015-11-30
Added checking layer in setPassword()

mmoreram:
* Covered with tests

### DIFF
--- a/src/Elcodi/Component/User/Entity/Abstracts/AbstractUser.php
+++ b/src/Elcodi/Component/User/Entity/Abstracts/AbstractUser.php
@@ -25,6 +25,7 @@ use Elcodi\Component\Core\Entity\Traits\EnabledTrait;
 use Elcodi\Component\Core\Entity\Traits\IdentifiableTrait;
 use Elcodi\Component\User\Entity\Interfaces\AbstractUserInterface;
 use Elcodi\Component\User\Entity\Traits\LastLoginTrait;
+use Elcodi\Component\User\Exception\InvalidPasswordException;
 
 /**
  * AbstractUser is the building block for simple User entities,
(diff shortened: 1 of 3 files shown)


mmoreram:
Related to https://github.com/elcodi/elcodi/pull/1013

src/Elcodi/Component/User/Entity/Abstracts/AbstractUser.php
```diff
@@ -311,10 +312,16 @@ public function getFullName()
      */
     public function setPassword($password)
     {
-        if (null !== $password) {
-            $this->password = $password;
+        if (null === $password) {
+            return $this;
         }
 
+        if (!is_string($password) || empty($password)) {
```
EmanueleMinotto: what about a `trim` instead of `empty`? this change should disallow passwords made of white spaces

mmoreram:
@EmanueleMinotto solved.
Anyway, think that in password is never saved the real password but the hash (in our case). These changes are for providing the component the consistency, but Bamboo uses a hasher for the password.
I would add a Behat feature in Bamboo for testing these scenarios.

EmanueleMinotto:
I know, but the User component could also be indipendent, so I always try to think if it's used indipendently from elcodi and/or bamboo (maybe too strange to use it outside of them? :\ I don't know...)

mmoreram:
@EmanueleMinotto yes, absolutely. Just saying that we should test this feature as well in Bamboo :)

EmanueleMinotto:
oh right :)

mmoreram:
@EmanueleMinotto feel free to thumbs up! :dancer:

EmanueleMinotto:
sorry for the missing question, but what about the same behaviour for `setPassword(null)` and `setPassword('')`? Currently are different

mmoreram:
Yes. when you pass null, is like updating nothing (is for forms).
When you pass empty, is like updating explicitly an empty string

EmanueleMinotto:
oh you're right :) well done than :+1:

PR #991, opened 2015-11-21
Testing

EmanueleMinotto:
I'm writing the missing tests.

New dependencies (development only):
* https://github.com/matthiasnoback/SymfonyDependencyInjectionTest : helps writing faster the tests related to DependencyInjection
* https://github.com/padraic/mockery : PHPUnit doesn't allow the creation of external mocks, instead this package can be used inside Twig tests, created following http://twig.sensiolabs.org/doc/advanced.html#testing-an-extension

Any other package that could improve tests or their creation is welcome. :)
I'm leaving this as a WIP because it's really big and I prefer to receive feedbacks while writing them.

## Bundles
- [ ] AttributeBundle
- [ ] BannerBundle
- [ ] CartBundle
- [ ] CartCouponBundle
- [ ] CommentBundle
- [ ] ConfigurationBundle
- [x] CoreBundle
- [ ] CouponBundle
- [ ] CurrencyBundle
- [ ] EntityTranslatorBundle
- [ ] FixturesBoosterBundle
- [ ] GeoBundle
- [ ] LanguageBundle
- [ ] MediaBundle
- [ ] MenuBundle
- [ ] MetricBundle
- [ ] NewsletterBundle
- [ ] PageBundle
- [ ] PaymentBundle
- [ ] PluginBundle
- [ ] ProductBundle
- [ ] RuleBundle
- [ ] ShippingBundle
- [ ] SitemapBundle
- [ ] StateTransitionMachineBundle
- [ ] StoreBundle
- [ ] TaxBundle
- [ ] UserBundle
- [ ] ZoneBundle

## Components
- [x] Attribute https://github.com/elcodi/elcodi/pull/1015
- [x] Banner https://github.com/elcodi/elcodi/pull/1028
- [x] Cart https://github.com/elcodi/elcodi/pull/1027
- [ ] CartCoupon
- [ ] Comment
- [ ] Configuration
- [ ] Core
- [ ] Coupon
- [ ] Currency
- [ ] EntityTranslator
- [ ] Geo
- [ ] Language
- [ ] Media
- [ ] Menu
- [ ] MetaData
- [ ] Metric
- [ ] Newsletter
- [ ] Page
- [ ] Payment
- [ ] Plugin
- [ ] Product
- [ ] Rule
- [ ] Shipping
- [ ] Sitemap
- [ ] StateTransitionMachine
- [ ] Store
- [x] Tax https://github.com/elcodi/elcodi/pull/1025
- [ ] User
- [ ] Zone

PR #1015, opened 2015-11-30
Tests for the Attribute component

EmanueleMinotto:
Unit tests covering the Attribute component.

Also some tests for the Core are included (because used inside the Attribute tests).

This is a split of https://github.com/elcodi/elcodi/pull/991

### DIFF
--- a/src/Elcodi/Component/Core/Tests/UnitTest/Entity/AbstractEntityTest.php
+++ b/src/Elcodi/Component/Core/Tests/UnitTest/Entity/AbstractEntityTest.php
@@ -1,0 +1,307 @@
+<?php
+
+/*
+ * This file is part of the Elcodi package.
+ *
+ * Copyright (c) 2014-2015 Elcodi Networks S.L.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * Feel free to edit as you please, and have fun.
+ *
+ * @author Marc Morera <yuhu@mmoreram.com>
+ * @author Aldo Chiecchia <zimage@tiscali.it>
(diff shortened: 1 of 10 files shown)


src/Elcodi/Component/Core/Tests/UnitTest/Entity/AbstractEntityTest.php
```diff
@@ -0,0 +1,312 @@
...
+        );
+
+        try {
+            $class->$setter(null);
+            $this->assertNull($class->$getter());
+        } catch (Exception $e) {
+            if ($field['nullable']) {
+                $this->fail("Method $setter of entity $classNamespace is not nullable");
+            }
+        }
```
EmanueleMinotto: imo the try/catch should be wrapped in the `if ($field['nullable']) {` instead

src/Elcodi/Component/Core/Tests/UnitTest/Entity/AbstractEntityTest.php
```diff
@@ -0,0 +1,312 @@
...
+            $value,
+            $class->$getter()
+        );
+
+        try {
+            $class->$setter(null);
+            $this->assertNull($class->$getter());
+        } catch (Exception $e) {
+            if ($field['nullable']) {
+                $this->fail("Method $setter of entity $classNamespace is not nullable");
```
EmanueleMinotto: What about a `sprintf('Method %s of entity %s is not nullable', $setter, $classNamespace)` here?

mmoreram:
I'm going to push a change to make all tests green. Can I squash both commits?

EmanueleMinotto:
@mmoreram sure :)

Issue #1016, opened 2015-12-05
Fix all yaml files where syntax is broken


mmoreram:
Symfony Yaml 2.8.0 has deprecated this syntax.

```yaml
some_service:
    class: ...
    arguments:
        - @whatever
``` 

because a scalar cannot be started by the reserved `@` indicator.
We must fix all files to that, with simple quotes:

```yaml
some_service:
    class: ...
    arguments:
        - '@whatever'
```

PR #1018, opened 2015-12-05
Quoted strings correctly to solve yaml component deprecation

EmanueleMinotto:
Closes https://github.com/elcodi/elcodi/issues/1016

> Starting an unquoted string with `@`, `` ` ``, `|`, or `>` leads to a `ParseException`.

Reference: https://github.com/symfony/symfony/blob/2.8/UPGRADE-3.0.md#yaml

### DIFF
--- a/src/Elcodi/Bundle/ConfigurationBundle/Resources/config/doctrine/Configuration.orm.yml
+++ b/src/Elcodi/Bundle/ConfigurationBundle/Resources/config/doctrine/Configuration.orm.yml
@@ -9,7 +9,7 @@ Elcodi\Component\Configuration\Entity\Configuration:
             length: 255
             nullable: false
         key:
-            column: `key`
+            column: '`key`'
             type: string
             length: 255
             nullable: false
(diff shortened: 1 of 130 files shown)


src/Elcodi/Bundle/ConfigurationBundle/Resources/config/doctrine/Configuration.orm.yml
```diff
@@ -9,7 +9,7 @@ Elcodi\Component\Configuration\Entity\Configuration:
             length: 255
             nullable: false
         key:
-            column: `key`
+            column: '`key`'
```
EmanueleMinotto: I'm not fully sure about this: no errors are given, but is it required?

mmoreram:
Bravo, Bravissimo.
This is not going to be tested properly, because yaml 3.0.0 is not being used.
Please, be sure that you're forcing yaml 3.0.0 in your localhost.

EmanueleMinotto:
schermata 2015-12-07 alle 00 01 50

Tested using `vendor/bin/phpunit`

mmoreram:
@EmanueleMinotto you're good man... you're good! :dancer:

PR #1028, opened 2015-12-08
Implemented tests for Banner component

EmanueleMinotto:
Unit tests covering the Banner component.

This is a split of #991

### DIFF
--- a/src/Elcodi/Component/Banner/Tests/UnitTest/Entity/BannerZoneTest.php
+++ b/src/Elcodi/Component/Banner/Tests/UnitTest/Entity/BannerZoneTest.php
@@ -1,0 +1,122 @@
+<?php
+
+/*
+ * This file is part of the Elcodi package.
+ *
+ * Copyright (c) 2014-2015 Elcodi Networks S.L.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
+ * Feel free to edit as you please, and have fun.
+ *
+ * @author Marc Morera <yuhu@mmoreram.com>
+ * @author Aldo Chiecchia <zimage@tiscali.it>
+ * @author Elcodi Team <tech@elcodi.com>
+ */
+
+namespace Elcodi\Component\Banner\Tests\UnitTest\Entity;
+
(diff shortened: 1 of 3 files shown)


src/Elcodi/Component/Banner/Tests/UnitTest/Services/BannerManagerTest.php
```diff
@@ -0,0 +1,62 @@
...
+ * @author Elcodi Team <tech@elcodi.com>
+ */
+
+namespace Elcodi\Component\Banner\Tests\UnitTest\Services;
+
+use Elcodi\Component\Banner\Repository\BannerRepository;
+use Elcodi\Component\Banner\Services\BannerManager;
+use PHPUnit_Framework_TestCase;
+
+class BannerManagerTest extends PHPUnit_Framework_TestCase
```
mmoreram: I don't understand the reason of this test

EmanueleMinotto: It's a classic unit test, just to test that `BannerManager` continue to have the same specification.
For example: `getBannersFromBannerZoneCode` shouldn't throw an exception, if it does (with a random string) than there's a bug.

src/Elcodi/Component/Banner/Tests/UnitTest/Entity/BannerZoneTest.php
```diff
@@ -0,0 +1,124 @@
...
+        ];
+    }
+
+    /**
+     * Test the string casting.
+     */
+    public function testToString()
+    {
+        $object = new BannerZone();
+        $name = sha1(rand());
```
mmoreram: why not "name"? both rand and sha1 use resources

src/Elcodi/Component/Banner/Tests/UnitTest/Services/BannerManagerTest.php
mmoreram: Well, in that case we could over this cases with functional tests

EmanueleMinotto:
@mmoreram removed the `BannerManagerTest` and updated the PR

mmoreram:
Awesome :)

PR #1034, opened 2015-12-11
Improved the abstract factory method

mmoreram:
* Because namespace is the primary key of the entity, using find, repository will make use
  of caching layer (entity instances layer).


EmanueleMinotto:
:+1:

Issue #1042, opened 2015-12-16
[Language] Deprecate LanguageExtension with global function


mmoreram:
Add a deprecation for LanguageExtension.
The final project should be responsible to define a global instead of Elcodi.
To be removed in 2.0.0

PR #1043, opened 2015-12-16
Issue #1042 Deprecated class and service

mmoreram:


src/Elcodi/Component/Language/Twig/LanguageExtension.php
```diff
@@ -19,9 +19,12 @@
 
 use Twig_Extension;
 use Twig_Extension_GlobalsInterface;
-
```
EmanueleMinotto: no more php-formatter?

mmoreram: LOL, yes, sorry :/
Let's merge this and will be solved in next PR, agree?
BTW, we need to find why php-cs-fixer is ignoring the configuration of php_cs file (is explicit that this check is disabled...)

EmanueleMinotto: :+1:

PR #1047, opened 2016-01-04
CI scripts improvements

EmanueleMinotto:
* a switch for the `preferred-install` option based on the `GH_TOKEN` env variable
* removed vendor and /tmp/Elcodi directories from travis cache (the first to avoid invalidating the cache because of Packagist metadata and the second to prevent old stored data)
* added nightly to the testing matrix (under the `allowed_failures` section)
* removed appveyor settings: clone_folder, platform and matrix because don't improve anything

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,33 +4,28 @@ sudo: false
 
 cache:
   directories:
-    - vendor
     - $HOME/.composer/cache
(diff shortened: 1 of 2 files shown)


EmanueleMinotto:
ping @mmoreram

PR #1048, opened 2016-01-07
Avoid saving translation if form is invalid

xphere:
When invalid data is entered in a form with `EntityTranslatorFormEventListener`, translations should not be saved. But they are.
This can lead to Doctrine exceptions about the entity being not saved (id is `null`) or other types of unexpected behaviour.


mmoreram:
Great! Thanks @xphere :D:D:D

PR #1049, opened 2016-01-07
Updated and Improved Purchasable Schema map

mmoreram:
* This PR is important in order to really work with a better purchasable design.
* Now everything is a purchasable object (pack, products, variant...) so all relations are done
  against this table
* Using doctrine inheritance (class) and updated all relations to work with them
* Now, for example, packs can have other packs, and some implementations are nicely complexity reduced

This PR adds important BC Break, not on implementation but in schema, so if this is accepted, then the
project should be tagged as 2.0.0 (what is not a problem at all... versions are for that...)


EmanueleMinotto:
Looks good to me :+1: 
But are you for the major change? There are still many things in the roadmap :\ http://elcodi.io/docs/book/roadmap/

PR #1053, opened 2016-01-11
Feature/updated purchasable schema

mmoreram:
**[Warning]** - This is the same as https://github.com/elcodi/elcodi/pull/1049 but redone to the right branch

* This PR is important in order to really work with a better purchasable design.
* Now everything is a purchasable object (pack, products, variant...) so all relations are done
  against this table
* Using doctrine inheritance (class) and updated all relations to work with them
* Now, for example, packs can have other packs, and some implementations are nicely complexity reduced

This PR adds important BC Break, not on implementation but in schema, so if this is accepted, then the
project should be tagged as 2.0.0 (what is not a problem at all... versions are for that...)


mmoreram:
ping @EmanueleMinotto 

This change requires a new major version. Of course, all versions are recomputed :)
Everything stays as before, but with a different version id.

PR #1056, opened 2016-01-18
Coupons Query Language (CQL)

mmoreram:
We have developed a coupons query language, a special query language for coupons design.
At the moment, only *NxM* coupons have been applied, and the query language only contains logic for
products, categories and manufacturers.

For example:

**3x2:p(13,34)|m(3)**

This means: buy 3, pay only 2, including all products of manufacturer 3 or products with id 13 and 34.
Here we have some extra examples

**2x1** - buy 2, pay 1, to all products
**4x3:c(13)** - buy 4, pay 3, to all products of category 13

This PR introduces as well modifiers. At the moment we have only 5 modifiers.
They can be combined.

**2x1:c(13):G**

*P* - Only products - When exists, only Products are applied
*V* - Only variants - When exists, only Variants are applied
*K* - Only packs - When exists, only Packs are applied
*G* - Group elements - When exists, coupons are applied to all elements matching the expression. This means that they are threated as the same
    element before applying the coupon. for example, if **2x1** for category 13 is applied with G modifier (**2x1:c(13):G**), all products, variants and packs with category 13 will match, they will be ordered from less expensive to more expensive, and the first half will be discounted. If the modifier is not introduced, all elements will be grouped by type and id and after this, **2x1** will be applied for each one.

If type modifiers are combined, then you can mix types

**2x1:c(13):PVG** - buy 2, pay 1, to all products with category 13 that are products, variants, grouping them all as one before applying.

This implementation is explicitely done for extension, with adapters and tags in Symfony DIC.

Documentation PR - https://github.com/elcodi/docs/pull/123
Documentation Web - http://elcodi.io/docs/book/coupon-architecture/#coupon-types

### DIFF
--- a/src/Elcodi/Bundle/CartCouponBundle/Resources/config/applicators.yml
+++ b/src/Elcodi/Bundle/CartCouponBundle/Resources/config/applicators.yml
@@ -1,0 +1,67 @@
+services:
+
+    #
+    # Applicators
+    #
+    elcodi.cart_coupon_applicator.absolute:
+        class: Elcodi\Component\CartCoupon\Applicator\AbsoluteCartCouponApplicator
+        arguments:
+            - '@elcodi.wrapper.currency'
+            - '@elcodi.converter.currency'
+        tags:
+            - { name: elcodi.cart_coupon_applicator }
+
+    elcodi.cart_coupon_applicator.mxn_abstract:
+        class: Elcodi\Component\CartCoupon\Applicator\Abstracts\AbstractMxNCartCouponApplicator
+        abstract: true
+        arguments:
+            - '@elcodi.wrapper.currency'
+            - '@elcodi.converter.currency'
+            - '@elcodi.cart_coupon_applicator_function_collector'
+
+    elcodi.cart_coupon_applicator.mxn_group:
+        class: Elcodi\Component\CartCoupon\Applicator\MxNGroupCartCouponApplicator
+        parent: elcodi.cart_coupon_applicator.mxn_abstract
+        tags:
+            - { name: elcodi.cart_coupon_applicator }
+
+    elcodi.cart_coupon_applicator.mxn_specific:
+        class: Elcodi\Component\CartCoupon\Applicator\MxNSpecificCartCouponApplicator
+        parent: elcodi.cart_coupon_applicator.mxn_abstract
+        tags:
+            - { name: elcodi.cart_coupon_applicator }
+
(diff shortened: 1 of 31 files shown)


EmanueleMinotto:
Uhm... seems really complex to me, there isn't anything we could do with symfony expressions? :\

mmoreram:
It works with symfony expressions :)
This is the core, of course, and in Bamboo we have to build a transformer to add this UX layer :)

mmoreram:
cc/ @xphere

src/Elcodi/Bundle/CartCouponBundle/Resources/config/applicators.yml
```diff
@@ -0,0 +1,67 @@
+services:
+
+    #
+    # Applicators
+    #
+    elcodi.cart_coupon_applicator.absolute:
+        class: Elcodi\Component\CartCoupon\Applicator\AbsoluteCartCouponApplicator
+        arguments:
+            - @elcodi.wrapper.currency
```
EmanueleMinotto: these should be wrapped with (single) quotes, the same below

mmoreram: Totally right... :) Thanks!

src/Elcodi/Component/Geo/Command/LocationLoadCommand.php
```diff
@@ -100,6 +100,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
                         'Country ' . $countryCode . ' loaded from repository'
                     );
             } catch (Exception $exception) {
+                echo $exception->getMessage();
+                die();
```
EmanueleMinotto: why this `echo` + `die`?

mmoreram: Because is a bug XD
I am reviewing the PR now as well :)

mmoreram:
I will rebase all commits before merging ;)

EmanueleMinotto:
Looks good to me :+1: but imo a chapter on elcodi/docs should be added asap after merging hehe

mmoreram:
@EmanueleMinotto okay! I will do it right now :)

Thanks!

mmoreram:
@EmanueleMinotto Documentation updated - http://elcodi.io/docs/book/coupon-architecture/#coupon-types

EmanueleMinotto:
well done! :+1:

Issue #981, opened 2015-11-17
[Plugin Component] Elcodi always call a plugin even is not enabled


manugarciaes:
If you deactivate a plugin with enabled = 0, elCodi always call this plugin and you can get a error or another problems, for example if you delete this plugin in your project from disk and change enabled = 0 (maybe is not the best example)  you get a error:

Attempted to load class "AtresmediahfMarketplaceShippingBundle" from namespace "Atresmediahf\Plugin\MarketplaceShippingBundle".
Did you forget a "use" statement for another namespace?

mmoreram:
@manugarciaes what do you mean by deactivating? Removing from Kernel?

miguelvilata:
This issue also affect the way that plugin routes are loaded dynamically by the custom router inside addPluginsRoutesCollection. I'm working in a exist() method in Plugin entity that returns true or false based on plugin files existence.

I think I could make a PR next monday

PR #1058, opened 2016-01-19
[Plugin] Check if plugin really exists before load routing information

mmoreram:
Related to #981

### DIFF
--- a/src/Elcodi/Component/Plugin/Exception/PluginNotFoundException.php
+++ b/src/Elcodi/Component/Plugin/Exception/PluginNotFoundException.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of the Elcodi package.
+ *
+ * Copyright (c) 2014-2016 Elcodi Networks S.L.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ *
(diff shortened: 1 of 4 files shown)


src/Elcodi/Component/Plugin/Exception/PluginNotFoundException.php
```diff
@@ -0,0 +1,38 @@
...
+class PluginNotFoundException extends RuntimeException
+{
+    /**
+     * @return PluginNotFoundException
+     */
+    public static function createPluginRouteNotFound(Plugin $plugin)
+    {
+        $message = sprintf('Elcodi cannot find plugin %s. Make sure it exists as a vendor by checking your composer, or load elcodi:install command in order to update the plugin database', $plugin->getNamespace());
+
+        return new self($message);
```
EmanueleMinotto: Since this is not declared as `final`, it could be extended, so here would be better to use `return new static($message)` ([ref](http://stackoverflow.com/a/5197655/871913))

mmoreram: In fact, should be declared as final... as all other extensions in Elcodi :)

Thanks for that detail!

EmanueleMinotto:
:+1:

PR #1059, opened 2016-01-21
Make country option mandatory for all command scripts

pramoddas:
1. Issue when command is executed "php app/console elcodi:locations:populate"
2. Made changes so that the above command throws an exception when --country  option is not specified...

### DIFF
--- a/src/Elcodi/Component/Geo/Command/LocationPopulateCommand.php
+++ b/src/Elcodi/Component/Geo/Command/LocationPopulateCommand.php
@@ -76,6 +76,8 @@ class LocationPopulateCommand extends AbstractLocationCommand
     {
         $this->startCommand($output, true);
         $countries = $input->getOption('country');
+        if (!$countries)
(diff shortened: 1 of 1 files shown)


EmanueleMinotto:
please run the coding style tools ([ref](https://github.com/elcodi/elcodi/blob/master/CONTRIBUTING.md))

mmoreram:
Thanks for that PR :D

This Exception should be thrown in all commands. @EmanueleMinotto are you agree?

@pramoddas please, style check should be applied :)

EmanueleMinotto:
sure @mmoreram, imo something more specific than an `\InvalidArgumentException`, but still no idea what exactly

mmoreram:
I'm afraid that an Exception will make people think that there is an issue with Elcodi, so maybe a message should be enough...

mmoreram:
Closing in favor of https://github.com/elcodi/elcodi/pull/1069

Issue #999, opened 2015-11-27
Strict User's gender value


EmanueleMinotto:
Currently the user's gender (both admin or normal user) can be everything because it's based on integers: https://github.com/elcodi/elcodi/blob/master/src/Elcodi/Component/User/Entity/Abstracts/AbstractUser.php#L170

Its value should be more strict.

mmoreram:
@EmanueleMinotto for example?

EmanueleMinotto:
I like a lot this approach: https://github.com/puli/cli/pull/23#discussion_r44212816

We could still use the integer (I don't know, maybe in future to add a "I don't want to say my sex"...) but using constants associated to integers different developers and data can't do mistakes.
What do you think?

mmoreram:
Can you tell me what sex values we should add? XD

EmanueleMinotto:
afaik there are only two in this world XD so imo something like this could be used:
* `const UNDEFINED = 0`
* `const MALE = 1`
* `const FEMALE = 2`

So the final developer must only use `setGender(User::MALE)`, more easy to understand than `setGender(1)`.

Why the undefined? Some users could decide to not show their sex, so this possibility is available for them.

This could also be useful for filter based on bitmasks, for example:
* unisex dresses: `User::MALE | User::FEMALE`
* what did the admin forget to mark: `User::UNDEFINED`

mmoreram:
@EmanueleMinotto okay then :D

mmoreram:
@EmanueleMinotto could you please address this?

EmanueleMinotto:
@mmoreram sure, I'll do it asap

mmoreram:
Thanks dude :D

PR #1062, opened 2016-01-24
Strict AbstractUser gender

EmanueleMinotto:
Closes https://github.com/elcodi/elcodi/issues/999

### DIFF
--- a/src/Elcodi/Component/User/Entity/Abstracts/AbstractUser.php
+++ b/src/Elcodi/Component/User/Entity/Abstracts/AbstractUser.php
@@ -39,6 +39,38 @@ abstract class AbstractUser implements AbstractUserInterface
         EnabledTrait;
 
     /**
+     * Gender not defined (automatically created user or privacy preferences).
+     *
+     * @var int
+     */
+    const GENDER_UNDEFINED = 0;
+
+    /**
+     * Gender male.
+     *
+     * @var int
+     */
+    const GENDER_MALE = 1;
+
+    /**
+     * Gender female.
+     *
+     * @var int
+     */
+    const GENDER_FEMALE = 2;
+
+    /**
+     * Allowed user genders.
+     * 
+     * @var array
+     */
+    private static $genders = [
+        self::GENDER_UNDEFINED,
+        self::GENDER_MALE,
+        self::GENDER_FEMALE,
+    ];
+
+    /**
      * @var string
      *
      * Password
@@ -74,11 +106,11 @@ abstract class AbstractUser implements AbstractUserInterface
     protected $lastname;
 
     /**
-     * @var int
+     * User gender, by default undefined.
      *
-     * gender
+     * @var int
      */
-    protected $gender;
+    protected $gender = self::GENDER_UNDEFINED;
 
     /**
      * @var DateTime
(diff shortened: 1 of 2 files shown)


src/Elcodi/Component/User/Entity/Abstracts/AbstractUser.php
```diff
@@ -38,6 +38,21 @@
         DateTimeTrait,
         EnabledTrait;
 
+    const GENDER_UNDEFINED = 0;
```
mmoreram: Undefined? Can someone have undefined sex? Shouldn't be this null? And being null, this wouldn't be necessary, right?

By default, null (undefined), and each factory should define the default value

src/Elcodi/Component/User/Entity/Abstracts/AbstractUser.php
```diff
@@ -38,6 +38,21 @@
...
+    const GENDER_MALE = 1;
```
mmoreram: We should add the PHPDoc (like other variables)

src/Elcodi/Component/User/Entity/Abstracts/AbstractUser.php
```diff
@@ -74,11 +89,11 @@
...
 
     /**
-     * @var int
+     * User gender, by default undefined.
      *
-     * gender
+     * @var int
      */
-    protected $gender;
+    protected $gender = self::GENDER_UNDEFINED;
```
mmoreram: Why do you change the order?

mmoreram:
Thanks @EmanueleMinotto 

We should consider these comments. Furthermore, the default gender should involve as well the factory

src/Elcodi/Component/User/Entity/Abstracts/AbstractUser.php
EmanueleMinotto: :+1:

src/Elcodi/Component/User/Entity/Abstracts/AbstractUser.php
EmanueleMinotto: The undefined value is used for two use cases:
1. https://github.com/elcodi/elcodi/issues/999#issuecomment-160705453 (last point)
2. if a user doesn't want to define their sex, don't know the reason exactly, but I know it's a common thing around the net

src/Elcodi/Component/User/Entity/Abstracts/AbstractUser.php
EmanueleMinotto: the constant is used here to keep the entity consistent, so there's no possibility to have a `null` or a `GENDER_UNDEFINED` value, at least there's the `GENDER_UNDEFINED` value.

EmanueleMinotto:
ping @mmoreram

mmoreram:
@EmanueleMinotto The factory should initialize with anything?

EmanueleMinotto:
I don't see a complex initialization for the `AbstractUser` nor something based on external data, so this attribute could be initializited using default attributes initialization (to keep the state too).
Also to force the initialization from factories, the constructor should be private.

I'm not sure about the factory in this PR :\ no?

mmoreram:
Okay then :)

Could you please rebase the PR?
As soon as is rebased and green, feel free to merge & close :)

Well done dude!

EmanueleMinotto:
@mmoreram rebased :)

PR #1069, opened 2016-02-09
Added country check in all geo commands

mmoreram:


mmoreram:
@EmanueleMinotto what do you think about this PR?

src/Elcodi/Component/Geo/Command/Abstracts/AbstractLocationCommand.php
```diff
@@ -72,6 +73,26 @@ protected function configure()
...
 
     /**
+     * Get a country list from an input object or throw an Exception if none is
+     * properly defined.
+     *
+     * @param InputInterface $input Input
+     *
+     * @return array Country array from input
+     *
+     * @throws InvalidArgumentException Countries not found
```
EmanueleMinotto: The *throw* annotation should be between *param* and *return*.

EmanueleMinotto:
Sorry, missed the notification, just a little thing about annotations position, but looks good to me :)
:+1:

PR #1072, opened 2016-02-16
Mixed improvements

EmanueleMinotto:
1. the `__DIR__` magic constant can be used since we are require PHP 5.4+ from the composer.json
2. `mt_rand` [is recommended](http://php.net/manual/en/function.mt-rand.php) instead of `rand`


mmoreram:
Excellent @EmanueleMinotto 

Thanks :D

